PR #324, opened 2012-11-12
Added multipart/form-data encoding.

xclydes:
I have added the ability to specify a request be sent in the multipart/form-data encoding, as well as adding a file to be sent as part of the request. 

 A number of other adjustments were also made so this form of encoding could be used for other purposes.
- The new FileParameter class takes a java.io.File which will be added as a part of the message body. This is an extension of the Parameter class.
- Each parameter can now have a disposition as specified by RFC-2388. 
  - It may be useful to add other attributes in the future such as content-type,  charset, content-transfer-encoding, etc.
- Individual parameter types can now indicate whether or not they are to be included as part of the base string. This is to account for instances where parameters are identified separately from files at the server end. 
  - An example of this is in PHP, where a single submission results in fields going to the $_POST array while uploads go the $_FILES array. This results in the signature expected by the server being different from that sent by the client.

It is showing that the entire file was modified, in the case of Parameter.java for example. This is due to the format command in eclipse.  Sorry for any confusion it may cause.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ scribe.iml
 # Binaries
 
 target
+bin
--- a/README.textile
+++ b/README.textile
@@ -65,7 +65,7 @@ You can pull scribe from a maven repository, just add this to your __pom.xml__ f
 @<dependency>@
   @<groupId>org.scribe</groupId>@
   @<artifactId>scribe</artifactId>@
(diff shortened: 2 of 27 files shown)


EddyVerbruggen:
Hi guys, I'm curious to learn why this PR eventually was never merged.

And btw, I <3 Scribe!

PR #470, opened 2014-03-11
Add methods getRequestToken and getAccessToken with RequestTuner parameter

msyben:
As discussed by email here is the modification I've done :

Add methods getRequestToken and getAccessToken with RequestTuner parameter in interface OAuthService

Implements those methods in OAuth20ServiceImpl

Call the tuner.tune(this) in the send method of the Request class earlier (with nullity check).


kullfar:

PR #478, opened 2014-04-04
allow whitespace around callback and urls - fixes #474

### LINKED ISSUE #474: The regex at `Preconditions` should accept whitespaces
That regex(URL_PATTERN) isnt allowing any spaces arround the url, we should trim it for the users or give a more specific error.

norbertpotocki:
fixes https://github.com/fernandezpablo85/scribe-java/issues/474


kullfar:

PR #484, opened 2014-04-25
Add an abstraction around the implementation of sending HTTP requests

nlalevee:
Used in an heavy multithreaded environment, we had issues with HttpUrlConnection. Using Apache HttpClient fixed our issues. Hence the suggested path, which introduce the abstraction of sending an HTTP request.
The API has been kept backward compatible, tests still pass.


kullfar:

PR #485, opened 2014-04-25
Add an abstraction around the implementation of sending HTTP requests, and an Apache HttpClient implementation

nlalevee:
Same as PR #484, but with an optional implementation via Apache HttpClient


kullfar:

PR #488, opened 2014-05-01
Updated request class to support an optional proxy

mnbogner:
Since Android does not pick up system proxy settings, I made these changes to allow the Scribe methods to connect through a Tor client.  The changes were necessary to support similar changes to a Flickr api library I'm working with, for which I am also submitting a pull request.


kullfar:

PR #500, opened 2014-07-18
Without the last slash, the url returned is not reachable.

etux:


kullfar:

PR #503, opened 2014-08-11
Introduce new ConnectionFactory/RequestFactory interfaces

dmitrygusev:
These interfaces used to create instances of `HttpURLConnection` and `OAuthRequest` respectively

I was trying to set custom [`SSLSocketFactory` for `HttpsURLConnection`](http://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HttpsURLConnection.html#setSSLSocketFactory(javax.net.ssl.SSLSocketFactory)) to setup certificate authentication while establishing secured connection, but found it's impossible using current implementation.

I tried using `RequestTuner` (getting `HttpsURLConnection` from request instance using reflection), but found it was too late for setting `SSLSocketFactory`, because connection was already opened by [`addBody(...)`](https://github.com/fernandezpablo85/scribe-java/blob/60aa4bb714a291a4694fc27085017ed83b00bbc3/src/main/java/org/scribe/model/Request.java#L114).

The change is implemented in a backward compatible way, so no existing clients should be affected.

The reason for two interfaces is that in Scribe's current implementation `Request` object is responsible for creating and configuring `HttpURLConnection`, so I decided to keep this behavior in favor of backwards compatibility and simply delegating call of `new URL(url).openConnection()` to new `ConnectionFactory#createConnection(url)`.

So in order to use custom `ConnectionFactory` you should implement custom `RequestFactory` interface where you create new `OAuthRequest` by passing this custom connection factory to `OAuthRequest`'s constructor.

Here's an example of using custom factories from real-life application: https://gist.github.com/dmitrygusev/388040ea136338bc8cea

### DIFF
--- a/src/main/java/org/scribe/model/OAuthConstants.java
+++ b/src/main/java/org/scribe/model/OAuthConstants.java
@@ -41,6 +41,7 @@ public class OAuthConstants
   public static final String HEADER = "Authorization";
   public static final Token EMPTY_TOKEN = new Token("", "");
   public static final String SCOPE = "scope";
+  public static final String SESSION_HANDLE = "oauth_session_handle";
 
   //OAuth 2.0
   public static final String ACCESS_TOKEN = "access_token";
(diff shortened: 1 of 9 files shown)


kullfar:

PR #517, opened 2014-10-22
BitBucket Builder Api, one Trello Bug

yuanbohan:
@scribe-java ers, this push is to support BitBucket api. Hope it will help.


kullfar:

PR #521, opened 2014-11-19
Deprecate Request#setConnectionKeepAlive()

wldaunfr:
See https://github.com/fernandezpablo85/scribe-java/pull/520 for details.

### DIFF
--- a/src/main/java/org/scribe/model/Request.java
+++ b/src/main/java/org/scribe/model/Request.java
@@ -351,9 +351,8 @@ public class Request
   /**
    * Sets whether the underlying Http Connection is persistent or not.
    *
-   * @deprecated since 1.3.7 - does nothing. Prior implementation did not just modify keep-alive for the underlying
-   *             connection of this {@code Request}, but implicitly for <b>all</b> Http connections of the JVM instance.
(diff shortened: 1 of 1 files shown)


rockihack:
Please remove `System.setProperty("http.keepAlive", ...);` soon. It's a nasty thing to modify global jvm settings. `OAuth10aServiceImpl.getRequestToken()` also uses `OAuthRequest` and keepAlive defaults to false. There is currently no way to modify this behavior and it results in large ssl overhead if you send many requests.

https://github.com/scribejava/scribejava/blob/04195020558ea17ba1c271b85ef40977055942c8/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth10aServiceImpl.java#L45

kullfar:

rockihack:
The issue is not outdated.
As long as you use `System.setProperty("http.keepAlive", ...);` it remains the same.

kullfar:
yep. you are right. reopening the issue

kullfar:
eventualy have merged ;-) Thx to all

kullfar:
fixed

PR #527, opened 2014-12-23
Add Withings API

lenniboy:
Thanks for your great work!


kullfar:

hafizudinjoe:
Anyone can update Withings API?

kullfar:
You can try. Don't think it will cause a lot of troubles in case you already have test app and user

PR #533, opened 2015-01-13
Updated Facebook URL's in relation to Facebook breaking changes.

markjamesbutler:
Upgraded Facebook API URL's in relation to implementation of [Facebook API Versioning](https://developers.facebook.com/docs/apps/upgrading?locale=en_GB#v1tov2)

Without tying the library to a specific version, breaking changes would happen when Facebook automatically update to their latest API's. [See Version Schedules) (https://developers.facebook.com/docs/apps/versions?locale=en_GB)

I've upgraded to the earliest possible version to avoid other regressions in this change.

### DIFF
--- a/src/main/java/org/scribe/builder/api/FacebookApi.java
+++ b/src/main/java/org/scribe/builder/api/FacebookApi.java
@@ -6,13 +6,13 @@ import org.scribe.utils.*;
 
 public class FacebookApi extends DefaultApi20
 {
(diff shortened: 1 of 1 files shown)


markjamesbutler:
Thanks for merging this.  Will you be releasing a new version to maven central soon?  I'd like to upgrade my webapp with the new changes asap.

Thanks again,

Mark

fernandezpablo85:
There are still no significant changes for crafting a new release, you can
always build from master and use a snapshot version of scribe if you don't
want to keep the scribe-related sources in your code.

PR #535, opened 2015-01-27
Added new Implementation for OAuth 2.0 auth header

yoprogramo:
Added auth header as new Implementation OAuth 2.0, included BiblioEteca.com service (OAuth1.0 and 2.0) and example (only OAuth 1.0)


kullfar:

PR #545, opened 2015-03-20
fix trello example

xudifsd:
if we don't specify callback, trello will always redirect us to oob

https://stackoverflow.com/questions/26458079/trello-oauth-failed-to-redirect-me-to-oauth-callback-and-keeping-redirect-me-to/29159348#29159348

### DIFF
--- a/src/test/java/org/scribe/examples/TrelloExample.java
+++ b/src/test/java/org/scribe/examples/TrelloExample.java
@@ -11,6 +11,7 @@ public class TrelloExample
 {
   private static final String API_KEY = "your_api_key";
   private static final String API_SECRET = "your_api_secret";
+  private static final String CALLBACK = "your_site";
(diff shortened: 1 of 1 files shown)


fernandezpablo85:
I'm going to copy **word-by-word** one of the first paragraphs of the Getting Started Guide (that I wrote not because I like writing docs, but to address common issues like this):

>The example uses OOB OAuth, if you want to pass a callbackUrl so that Twitter redirects you there just  add a callback("http://your_callback_url") call before build()

xudifsd:
oh, sorry about that, I just jumped directly to the example, and thought example provide all I need.

fernandezpablo85:
The example provides a working example. Obviously it can't possibly provide
what everybody needs for all situations or there would be a lot more of
them.

xudifsd:
ok, thanks anyway, it's an awesome library

fernandezpablo85:
No problem! Thank you!

PR #555, opened 2015-06-01
LinkedIn OAuth 2.0 implementation

Vladislav-Zolotaryov:
LinkedIn OAuth 2.0 implementation
Addition of state field for CSRF protection (known to work with LinkedIn)
Example how to use LinkedIn OAuth 2.0

### DIFF
--- a/src/main/java/org/scribe/builder/ServiceBuilder.java
+++ b/src/main/java/org/scribe/builder/ServiceBuilder.java
@@ -21,6 +21,7 @@ public class ServiceBuilder
   private String callback;
   private Api api;
   private String scope;
+  private String state;
   private SignatureType signatureType;
   private OutputStream debugStream;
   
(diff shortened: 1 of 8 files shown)


simtse:
@fernandezpablo85 Could you have a look at this pull request and see if it can make it into the next release soon?

fernandezpablo85:
Oauth 2.0 support is discontinued.

xberger:
When I encountered the missing state verification I was puzzled to find the statement "Oauth 2.0 support is discontinued" buried just here, but not visible on the main readme, let alone in the example or service classes.
I would propose to clearly state the discontinuation on the main readme so that people can notice on time that important features are missing. Does that makes sense?

kullfar:
Why do you think OAuth2 LinkedIn is discontinued in ScribeJava?

This Pull Request and comments here are just history. Actual information is in the main readme and in the examples
https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/LinkedIn20Example.java

PR #556, opened 2015-06-02
adding support to the state parameter

leonardocordeiro:
Support to the state parameter for preventing cross-site request forgery

### DIFF
--- a/src/main/java/org/scribe/builder/ServiceBuilder.java
+++ b/src/main/java/org/scribe/builder/ServiceBuilder.java
@@ -1,6 +1,7 @@
 package org.scribe.builder;
 
 import java.io.*;
+
 import org.scribe.builder.api.*;
 import org.scribe.exceptions.*;
 import org.scribe.model.*;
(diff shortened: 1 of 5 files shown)


fernandezpablo85:
Can't you use a `RequestTuner` for this?

leonardocordeiro:
Hi Pablo, thanks for you reply.

The state parameter must be sent in Authorization URL for the Authorization Server and returned by this back to the Client Application.

I'm not sure if I got the idea exactly, but I do not see how can I use ```RequestTuner``` for it.

fernandezpablo85:
Indeed, my bad I thought the problem was in the `getAccessToken` url.

In this case is even easier:

```java
String authorizationUrl = service.getAuthorizationUrl(EMPTY_TOKEN) + "&state=" + "anyState";
```

I don't think we need an special case for it.

simtse:
@fernandezpablo85 , I have done the same as you have mentioned by hardcoding the `state` argument and it works. However, wouldn't it have more safety for the developer to specify the `state` String that they want? 

For LinkedIn OAuth 2.0, I also had to pass the `state` and I hard coded it to get it working. If scribe-java were to support LinkedIn OAuth 2.0 out-of-the-box, then this Pull request will allow the developers to not use a hard coded value in this library that could be easily exploited.

kullfar:

PR #559, opened 2015-07-08
super tiny improvement to maven pom changes

alph486:


Vladislav-Zolotaryov:
Really?

alph486:
Easy...Just for copy paste man, you don't have to merge it.

fernandezpablo85:
This would assume that scribe is your only dependency (and that's hardly the common case).

alph486:
Yes context was bad after looking sorry; I would have done something differently but clearly it's a bit too tedious. Thanks for looking.

fernandezpablo85:
no problem man, thanks for trying to improve scribe :)

PR #561, opened 2015-07-30
add getParameter method for easier query parameter access

akoskm:
I have to access the `screen_name` after receiving an access token. I think the `Token` class would be a perfect place for such a helper method.


fernandezpablo85:
LGTM

PR #572, opened 2015-10-05
Send the params in a url-encoded form in case of POST or PUT request

arribeth:
The 4.1.3 section of the rfc6749 specifies that the grant_type, code, redirect_uri and client_id parameters must be sent using the "application/x-www-form-urlencoded" format which implies the use of a url-encoded form in the body of the request for the POST and PUT methods.


kullfar:

PR #590, opened 2015-12-17
add a getter for GoogleToken.openIdToken.

chenjianjx:
I need to send it to my backend for openId validation.


scribejava-apis/src/main/java/com/github/scribejava/apis/google/GoogleToken.java
```diff
@@ -17,6 +17,10 @@ public GoogleToken(final String token, final String secret, final String rawResp
         super(token, secret, rawResponse);
         this.openIdToken = openIdToken;
     }
+    
+    public String getOpenIdToken() {
+		return openIdToken;
+	}
```
kullfar: do not use tabs pls ;-)
I'll reformat a bit your commit and will push it to master

Issue #596, opened 2016-01-07
refresh tokens


pdtyreus:
More an more OAuth 2 services are using "refresh" tokens to extend the life of access tokens. The `refresh_token` is usually returned along with the `access_token`, but scribe currently does not parse it out or store it in the `Token` object.

I see that you previously rejected a [pull request](https://github.com/scribejava/scribejava/pull/253) to bring refresh tokens in to this library. However, this was many years ago and I think that use of refresh tokens is much more common now. I've run across it in three major services already.

 * [Google Refresh Tokens](https://developers.google.com/identity/protocols/OAuth2WebServer?hl=en#offline)
 * [Facebook Refresh Tokens](https://developers.facebook.com/docs/facebook-login/access-tokens/expiration-and-extension)
 * [Medium Refresh Tokens](https://github.com/Medium/medium-api-docs)

With Google, it's especially important since the access tokens only live for an hour.

I'm volunteering to do this work as a pull request but wanted to check in with you before I do it. Can I add a `refreshToken` member to `Token`? Or extend it to a `RefreshableToken`? What type of modification would you accept as a PR?

kullfar:
Hi!
It's a very nice proposal.
Will be very happy to get such Pull Request. As for preferable type of modification, it's hard to understand now, what will be the best decision. Any modification, that will improve the library are welcome ;-)

Thanks,
Stas

kullfar:
done

PR #600, opened 2016-01-14
Support OAuth2 refresh tokens

pdtyreus:
The existing code was using the same base class for OAuth1 and OAuth2 request and access tokens. While nice for the API, this was a bit limiting for functionality. It was also confusing because OAuth2 does not have request tokens, and OAuth2 access tokens do not hold a secret. Since the secret was a required field in the token, we ended up passing in a lot of empty strings in places where users may not know this was correct to do.

I created the concept of RequestTokens and AccessTokens and refactored the code to be able to use the correct type depending on the API.

I modified the `OAuthService` class quite a bit to support the new version-specific functionality. However, *no existing APIs were broken.* I took pains to retain backwards compatibility even though it makes the API a bit uglier.

There are some new `Token` types:

 * `RequestToken` and `AccessToken` are interfaces for the service to use.
 * `OAuth1Token` - basically the same as the previous `Token` class. Only applies to OAuth1. Implements both `RequestToken` and `AccessToken`.
 * `OAuth2AccessToken` - new class that holds OAuth 2 concerns like `refresh_token`, `expires_in`, and `token_type`. Implements `AccessToken`.

I introduced a few new classes to help with parsing these tokens

 * `OAuth1RequestTokenExtractorImpl`
 * `OAuth2AccessTokenExtractorImpl`
 * `OAuth2JsonAccessTokenExtractorImpl`

All test cases pass and all the examples still run without changes.

The `OAuth20ServiceImpl` class now has a new method 
```
public OAuth2AccessToken refreshOAuth2AccessToken(final OAuth2AccessToken refreshToken)
```
that will exchange a short-lived token for a longer one. *I modified the Facebook and Google20 examples to show how to use the new refresh token functionality.*

Right now the only way to get an instance of an `OAuth20ServiceImpl` is to cast the response from `ServiceBuilder.build()`. We may want to consider adding methods like `.buildOAuth1Service()` or `.buildOAuth2Service()` that would automatically validate the type of the API and not require the end user to cast.

This is currently working for us as-in, but I'm very open to suggestions for improvement.

see issue #596


kullfar:
Great work! Thanks. I'll look closer a bit later

kullfar:
As for "I took pains to retain backwards compatibility even though it makes the API a bit uglier."
We can make it smoothly. I mean we can drop backward compatibility in the next version. And we can mark as "deprecated" "ugly" things in this version.

pdtyreus:
Great, thanks. I originally marked a bunch of methods in `OAuthService` as `@Deprecated` but it was giving a lot of compiler warnings for the examples, so I took that out. I'd love to work with you to figure out an elegant path forward for the library.

I think it would mostly come down to removing some methods that are not common between OAuth1 and OAuth2 from the `OAuthService` and then perhaps modifying the `ServiceBuilder` to handle the OAuth1 and OAuth2 cases slightly differently or doing something clever with generics.

kullfar:
Hi. I have prepared Pull Request to resolve this question
```
Right now the only way to get an instance of an OAuth20ServiceImpl is to cast the response from ServiceBuilder.build(). We may want to consider adding methods like .buildOAuth1Service() or .buildOAuth2Service() that would automatically validate the type of the API and not require the end user to cast.
````
https://github.com/scribejava/scribejava/pull/604

User will not need to choose the name of the method. He well pass API in the build() method and will get OAuth1Service or OAuth2Service based on the provided API (passed as argument to the build() method). As for me, it looks pretty well. And no need to casting

**Comment by pdtyreus** [comment]
Oh, yes, that's perfect!

Looks like there are a few conflicts with this branch, but once you merge  #604 into master I will rebase and get it ready again.

**Comment by kullfar** [comment]
Nice to hear this ;-) So. I have no barriers to merge it to the master. Have just done that.

**Comment by pdtyreus** [comment]
Great, everything is merged. There are no conflicts remaining with the upstream master. I double checked a few of the examples and things are working well. Both the Facebook and Google 2.0 examples are modified so that they attempt (and succeed) in refreshing the access tokens.

File: scribejava-apis/src/main/java/com/github/scribejava/apis/FacebookApi.java [hunk 1]
```diff
@@ -42,4 +52,32 @@ public String getAuthorizationUrl(final OAuthConfig config) {
         }
         return sb.toString();
     }
+
+    @Override
+    public OAuth20Service createService(OAuthConfig config) {
+        return new FacebookOAuth20ServiceImpl(this, config);
+    }
+    
+    private static class FacebookOAuth20ServiceImpl extends OAuth20Service {
```
kullfar: why private and not in the separated file?

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Google20Example.java
```diff
@@ -15,28 +15,31 @@
...
+        System.out.print(">>");
+        final String clientSecret = in.nextLine();
+        final String secretState = "state_" + new Random().nextInt(999_999);
         final OAuth20Service service = new ServiceBuilder()
                 .apiKey(clientId)
                 .apiSecret(clientSecret)
                 .scope("profile") // replace with desired scope
                 .state(secretState)
-                .callback("http://example.com/callback")
+                //.callback("http://example.com/callback")
```
kullfar: Why do we need no callback here?

scribejava-core/src/main/java/com/github/scribejava/core/model/Token.java
```diff
@@ -8,101 +8,7 @@
...
-        return token.equals(that.getToken()) && secret.equals(that.getSecret());
-    }
-
-    @Override
-    public int hashCode() {
-        return 31 * token.hashCode() + secret.hashCode();
-    }
+public interface Token extends Serializable {
+    public String getToken();
+    public boolean isEmpty();
```
kullfar: Do wee need isEmpty() here? Seems it's needed only in OAuth10aService

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth10aService.java
```diff
@@ -125,33 +193,46 @@ public void signRequest(final Token token, final AbstractRequest request) {
...
-    private String getSignature(final AbstractRequest request, final Token token) {
+    /**
+     * OAuth 2 APIs are not implemented and will throw an UnsupportedOperationException.
+     * 
+     * @return throws exception
+     */
+    @Override
+    public String getOAuth2AuthorizationUrl() {
+        throw new UnsupportedOperationException("OAuth 2 authorization urls are not supported for OAuth 1 APIs");
+    }
```
kullfar: We have separated OAuth10aService and OAuth20Service. Users are getting specific version after invoking build(). 
I thinks there is no need to continue to support common OAuthService interface in case of methods like this with "throw new UnsupportedOperationException()" in one of the Implementation.

kullfar: and with getOAuth1AuthorizationUrl() / getOAuth2AuthorizationUrl() names.
OAuth1/OAuth2 is in the class name already. I'll change tomorrow OAuthService, will try to get rid of getRequestToken method as an example.

kullfar:
I've just merged to the master commit https://github.com/scribejava/scribejava/commit/7b4475fca71706952564acb14d8aebf52cf4a4c8
I have moved method getRequestToken from OAuthService to OAuth10aService. And have maden methods getAuthorizationUrl, getAccessToken[Async] OAuth1 and OAuth2 specific (with Request Token to the OAuth1 and without Request Token to OAuth2). No more strange null-tokens in OAuth2 examples.
build() gives us a concrete OAuth version Service. And we need version-specific process of retrieving accesstoken, so there is no need to have common OAuthService abstraction for this purposes.

kullfar:
Please, rebase (not merge). It's hard to review changes at once, and hard to review old-master related commits. And try to fixup commits like "cleaning up new file headers" https://github.com/pdtyreus/scribejava/commit/813ac1639d92e28a22de6019efc3a165fee15411

scribejava-apis/src/main/java/com/github/scribejava/apis/FacebookApi.java
pdtyreus: I made it a private class because the overridden method returns an abstract class and the concrete implementation is not used or needed anywhere else in the code. So it was just an attempt to keep the code organized and encapsulated. I can easily change that if you prefer.

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Google20Example.java
pdtyreus: It was just easier to test as an OOB call. We can add the callback back if you prefer to have it that way.

scribejava-core/src/main/java/com/github/scribejava/core/model/Token.java
pdtyreus: You are correct. It's only needed for `signRequest` in `OAuth10aService`. Would you like me to take it out of the interface?

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth10aService.java
pdtyreus: Agreed

pdtyreus:
Thanks, I'll do a better job rebasing next time. This got a little more complicated than usual because there was so much work happening on master at the same time.

Let me know if there is something you would like me to look at or if you are handling the rest.

scribejava-apis/src/main/java/com/github/scribejava/apis/FacebookApi.java
```diff
@@ -42,4 +52,32 @@ public String getAuthorizationUrl(final OAuthConfig config) {
...
+
+        public FacebookOAuth20ServiceImpl(DefaultApi20 api, OAuthConfig config) {
+            super(api, config);
+        }
+
+        public OAuth2AccessToken refreshOAuth2AccessToken(final OAuth2AccessToken refreshToken) {
```
kullfar: BTW add \@Override annotation here

scribejava-apis/src/main/java/com/github/scribejava/apis/FacebookApi.java
kullfar: Yep. It's better to have separate file with service extending OAuth20Service. As it is with all other APIs and Services

scribejava-core/src/main/java/com/github/scribejava/core/model/RequestToken.java
```diff
@@ -0,0 +1,11 @@
+package com.github.scribejava.core.model;
+
+import com.github.scribejava.core.utils.Preconditions;
+
+/**
+ *
+ * @author Daniel Tyreus
+ */
+public interface RequestToken extends Token{
+    public String getSecret();
```
kullfar: public is redundant

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Google20Example.java
kullfar: Isn't it better to specify smth like "urn:ietf:wg:oauth:20:oob" here in such case?
And anyway, some comment like 'put here your OOB or callback url' will be good.

scribejava-core/src/main/java/com/github/scribejava/core/model/Token.java
kullfar: It sounds logical

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth10aService.java
kullfar: Have done this ;-)

kullfar:
Could you rebase this pull request (not merge)?
Or even create the new one. As you mentioned, there was a lot of work done in master, Pull request should become smaller. and without a lot of UnsupportedExceptions and a bucket of getOAuth[1|2]Token() methods

kullfar:
Thanks in advance and for already done job, which inspired changes, already in master branch

pdtyreus:
No problem. The rebasing is taking a little time because there is a lot of hand editing required. Once I'm done I will close this pull request and open a new one against the rebased branch.

PR #606, opened 2016-01-25
introduce checkstyle here

kullfar:
It's a public project, many people write here. We should try to have common style in sources. + Reduce review time by automatically checking some typical base mistakes/typos by maven-checkstyle plugin.


igaranina:
доне!)

PR #614, opened 2016-02-01
Stack Exchange authentication via OAuth 2.0

MichalFoksa:
Plus support response in gzip (Content-Encoding: gzip).

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/StackExchangeApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/StackExchangeApi.java
@@ -19,7 +19,7 @@ public class StackExchangeApi extends DefaultApi20 {
     private static final String AUTHORIZE_URL
     = "https://stackexchange.com/oauth?client_id=%s&redirect_uri=%s";
 
-    private StackExchangeApi() {
+    protected StackExchangeApi() {
     }
 
     private static class InstanceHolder {
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/StackExchangeApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/StackExchangeApi.java
@@ -1,0 +1,59 @@
+package com.github.scribejava.apis;
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.github.scribejava.core.model.OAuthConfig;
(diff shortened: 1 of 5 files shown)


kullfar:
I have cherry-picked your commit with gzip support to the master branch (with small fixes).
Rebase please this Pull request and provide me test user and test application credential, so I can run the example by myself

kullfar:
Thanks!

scribejava-apis/src/main/java/com/github/scribejava/apis/StackExchangeApi.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @author Michal Foksa
+ *
+ */
+public class StackExchangeApi extends DefaultApi20 {
+
+    private static final String AUTHORIZE_URL
+    = "https://stackexchange.com/oauth?client_id=%s&redirect_uri=%s";
+
+    private StackExchangeApi() {
```
kullfar: make it protected. It was my fault to make all API constructors private. I fixed that. In master branch all API constructors are protected, so anyone can extend API in their code if they want. But couldn't instantiate directly, forcing them to use singleton

kullfar:
the rest of code is ok (just make protected constructor please)

MichalFoksa:
Here you are, protected constructor.

MichalFoksa:
I can create a temporary test application, but I cannot create a test account there.

kullfar:
why not?

kullfar:
You can create test email and register new user anywhere, right?

MichalFoksa:
I would delete test app after the PR merge, but I cannot delete an account, and creating single purpose account is nasty. Actually having an account on `stackoverflow.com` might be beneficial for you in future - if you have to test the `StackExchangeExample.java`.

kullfar:
It's not secure to give permissions to access my real account for an unknown test application ;-)
FB have such feature as special temporary test account for a application. Nice decision I think.
If stackoverflow doesn't have such feature, we have no chance to not create single purpose temporary account ;-(

MichalFoksa:
An email address to send you test data, PLS.

kullfar:
Use this one pls
https://github.com/scribejava/scribejava/blob/master/pom.xml#L48

kullfar:
I have merged this Pull Request via console (not by github.com). So closing it here

PR #635, opened 2016-03-07
resource owner password credentials grant

maksimlikharev:
#1. Unification of async and sync invocations
#2. Resource owner grant, per https://tools.ietf.org/html/rfc6749#section-4.3


scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
```diff
@@ -32,9 +39,11 @@ public OAuth20Service(DefaultApi20 api, OAuthConfig config) {
     }
 
     public final OAuth2AccessToken getAccessToken(Verifier verifier) {
-        final Response response = createAccessTokenRequest(verifier,
-                new OAuthRequest(api.getAccessTokenVerb(), api.getAccessTokenEndpoint(), this)).send();
-        return api.getAccessTokenExtractor().extract(response.getBody());
+        return Uninterruptibles.getUninterruptibly( getAccessTokenAsync(verifier, null) );
```
kullfar: Are you sure it will work without ning http client in the classpath?

maksimlikharev:
of course it will not, but why you have async call without providing dependency for that?

kullfar:
it's optional dependency https://github.com/scribejava/scribejava/blob/master/pom.xml#L94
If you are not needed async http requests (as the most of ScribeJava users), you will not be forced to have ning http client in your classpath.

kullfar:
and without ning http client in your classpath you can use getAccessToken()

kullfar:
and to use async version of getAccessToken, you will have to add ning http client to the classpath.

maksimlikharev:
which makes 2 entry points for the http transport, complicates mocking and unit testing, provides different behavior in different conditions, that is async and sync request.

maksimlikharev:
and of course code duplication

maksimlikharev:
I'll change it back to separate flows, but this is very strange decision to have separate http flows

kullfar:
Yep. Async and sync is very different.
Sync - simple and common.
Async - if you really need it.
You have to choose what kind of http to use. Someone really needs async, but it's bad idea to force 'simple' user to learn complicated async and use ning http client.
Do you know better way how to achieve both goals?

As for mocking, I don't see any really big problems here.

maksimlikharev:
point is, they don't have to be different, differences are bad on so many levels, like for example: https://github.com/scribejava/scribejava/pull/638, exactly because sync and async differences. Anyway this is theological discussion at this point :)

maksimlikharev:
restored sync and async as separate flows

kullfar:
When we will add another async http client provider, we will have third branch of code ;-)

scribejava-core/src/main/java/com/github/scribejava/core/utils/Uninterruptibles.java
```diff
@@ -0,0 +1,37 @@
+package com.github.scribejava.core.utils;
+
+import com.github.scribejava.core.exceptions.OAuthException;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+public final class Uninterruptibles {
```
kullfar: We do not need this class anymore

scribejava-core/src/test/java/com/github/scribejava/core/oauth/CompletedFuture.java
```diff
@@ -0,0 +1,43 @@
+package com.github.scribejava.core.oauth;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ */
+class CompletedFuture<V> implements Future<V> {
```
kullfar: and this one as well

scribejava-core/src/main/java/com/github/scribejava/core/utils/Uninterruptibles.java
maksimlikharev: yes, will remove

scribejava-core/src/test/java/com/github/scribejava/core/oauth/CompletedFuture.java
maksimlikharev: no, look at unit test

maksimlikharev:
> When we will add another async http client provider, we will have third branch of code ;-)
You already have branch between sync/async, and my point previously exactly about that, so you don't need specific sync and async branch. Using async for both flows will simplify and unify code. Also it will allow pluggable async http client.

scribejava-core/src/test/java/com/github/scribejava/core/oauth/CompletedFuture.java
kullfar: Ah, yes. I see

kullfar:
With unified code 'simple' users will have to learn how to deal with ning http client, right?
And in case of pluggable async http client we still stick to multiple code branches, won't we?
Every http client will have their way of setting headers e.g.

maksimlikharev:
#1. simple users won't notice anything except extra dependency, ning which in turn would bring netty.  Netty is a core of any communication library nowadays anyway.

#2 yes, would require code refactoring to properly abstract http core, but technically netty as good as it gets and used in many mission critical applications now. 

Anyway, something for you to think about :)

maksimlikharev:
ok, dead code removed

kullfar:
Did you try to run
"mvn clean install"? ;-)
"You have 86 Checkstyle violations."

kullfar:
I have fixed violations, and merged your PR. with some changes. Will test and push the master here soon

kullfar:
done. in master

kullfar:
BTW, it will be great to have any XXXExample.java on using this flow

maksimlikharev:
I'll produce example later and it needs more unit tests for the other OAuth flows, btw checkstyle violations, do you have editorconfig?

kullfar:
thanks in advance. unit test is almost always good ;-)
no, I do not have editorconfig

PR #643, opened 2016-03-18
service builder more generic

andretietz:
This PR does not change anything functionality wise, but removes the second method from the service builder.

Take it, if you like it


kullfar:
Hm. nice improvement. Thanks! have merged

PR #645, opened 2016-03-21
Make API and Core proper OSGi artifacts.

jawi:
Adding the Felix Maven bundle plugin allows the resulting artifacts to be used as-is in an OSGi environment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,7 +198,7 @@
             </plugin>
         </plugins>
     </build>
-  
+
     <profiles>
         <profile>
             <id>release-sign-artifacts</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -224,8 +224,30 @@
...
+                        <executions>
+                            <execution>
+                                <id>bundle-manifest</id>
+                                <phase>process-classes</phase>
+                                <goals>
+                                    <goal>manifest</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
```
kullfar: Why do we need this plugin only on release-sign-artifacts profile?

jawi: Oh shoot, I've messed this one up. Should be a "general" build plugin, not a profile specific one. 

The reason for having this plugin is to prevent a change of artifact types as the Maven Bnd plugin by default only kicks in for "bundle" artifact types...

kullfar: Have just commited
https://github.com/scribejava/scribejava/commit/1dc3923153be6395c146cd8a17ff8f556cdeb225

kullfar:
merged with fixes to the master
https://github.com/scribejava/scribejava/commit/1dc3923153be6395c146cd8a17ff8f556cdeb225

PR #646, opened 2016-03-23
Add support for OpenID connect providers

jawi:
I've added some basic support for working with OpenID connect providers, such as Google and Microsoft. In contrast to the existing APIs, I've chosen to make this API a little more flexible in the sense that you can provide it the exact configuration it should use. This configuration can be obtained, for example, through the OpenID connect discovery mechanism.
Although there are no tests in this PR, I've tested it in another project against both Google and Microsoft.


kullfar:
Oh.. Interesting... OpenId isnot the goal for ScribeJava, but in case it won't make using ScribeJava more difficult... Thanks for PR. Will think on it

jawi:
The reason I've done this is that OpenID connect bolts on top of OAuth2 and well supported by ScribeJava. It saves me from writing lots of boilerplate code with the APIs provided by ScribeJava ;)

kullfar:
nice to hear, that ScribeJava helps ;-)
Yep, I know OpenID a bit, but not so good yet. Will read this weekend

NinoFloris:
I would love something like this to be merged into here, OIDC is well established by now and of great use.

kullfar:
You're definitely right! I have some plans, but a lack of time is my enemy.
I hope to merge something in medium term time.
Actually it's my number first of major improvements.

PR #649, opened 2016-03-31
Odnoklassniki Example improvements

thefilin:
Small improvements for Odnoklassniki Example

### DIFF
--- a/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
+++ b/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
@@ -18,14 +18,14 @@ public abstract class OdnoklassnikiExample {
 
     public static void main(String... args) {
         // Replace these with your client id and secret
-        final String clientId = "your client id";
-        final String publicKey = "your api secret";
-        final String clientSecret = "your client secret";
+        final String clientId  = "your api client id";
+        final String publicKey = "your api public key";
+        final String secretKey = "your api secret key";
 
         final OAuth20Service service = new ServiceBuilder()
                 .apiKey(clientId)
-                .apiSecret(clientSecret)
-                .scope("VALUABLE ACCESS")
+                .apiSecret(secretKey)
+                .scope("VALUABLE_ACCESS")
                 .grantType(OAuthConstants.AUTHORIZATION_CODE)
                 .callback("http://your.site.com/callback")
                 .build(OdnoklassnikiApi.instance());
--- a/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
+++ b/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
@@ -25,7 +25,6 @@ public abstract class OdnoklassnikiExample {
         final OAuth20Service service = new ServiceBuilder()
                 .apiKey(clientId)
                 .apiSecret(secretKey)
-                .scope("VALUABLE_ACCESS")
                 .grantType(OAuthConstants.AUTHORIZATION_CODE)
                 .callback("http://your.site.com/callback")
                 .build(OdnoklassnikiApi.instance());


scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
```diff
@@ -18,14 +18,14 @@
 
     public static void main(String... args) {
         // Replace these with your client id and secret
-        final String clientId = "your client id";
-        final String publicKey = "your api secret";
-        final String clientSecret = "your client secret";
```
kullfar: Why? What is secretKey in terms of OAuth2? There is client_secret, but no secret_key.

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
```diff
@@ -18,14 +18,14 @@
...
+        final String clientId  = "your api client id";
+        final String publicKey = "your api public key";
+        final String secretKey = "your api secret key";
 
         final OAuth20Service service = new ServiceBuilder()
                 .apiKey(clientId)
-                .apiSecret(clientSecret)
-                .scope("VALUABLE ACCESS")
+                .apiSecret(secretKey)
+                .scope("VALUABLE_ACCESS")
```
kullfar: May be better remove at all, because
"VALUABLE_ACCESS - access to all API methods, except users.getLoggedInUser and users.getCurrentUser. To have effect permission must also be set by Odnoklassniki side as well. To request permission send an email with id and shortname of your application to the address api-support@odnoklassniki.ru"

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
```diff
@@ -18,14 +18,13 @@
 
     public static void main(String... args) {
         // Replace these with your client id and secret
-        final String clientId = "your client id";
-        final String publicKey = "your api secret";
-        final String clientSecret = "your client secret";
+        final String clientId  = "your api client id";
+        final String publicKey = "your api public key";
```
kullfar: BTW, Seems like it's OK-specific stuff. And it's correct name  is 'applicationKey', right?

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
thefilin: I agree.
Removed.

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
kullfar: Ok, agreed with this change ;-)
according to OK docs, OAuth 'client_secret' is OK 'application secret key'.
And here we should provide data in OK terms. mapping to OAuth terms will be on the next lines
```
.apiSecret(secretKey)
````

thefilin: We don't modify part of library working with OAuth2.

But for using example developer should understand where to get these keys and we can use here terms of Odnoklassniki API.

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/OdnoklassnikiExample.java
thefilin: Yes, but some developers in first time can't understand what kind of key we want to receive.
Explicit name can resolve this mistake.

kullfar: I see. ok

thefilin:
Added link to the library in OK API documentation.

kullfar:
Great! thanks

kullfar:
tested, merged to the master

PR #657, opened 2016-04-29
added Keycloak OAuth2 template

jcon321:
Just wanted to provide a Keycloak template, I realize it is a little different than the other public OAuth servers since this is intended for a private server, but thought it could help.

thanks

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi20.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi20.java
@@ -1,0 +1,57 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package com.github.scribejava.apis;
+
+import com.github.scribejava.apis.service.KeycloakOAuthServiceImpl;
+import com.github.scribejava.core.builder.api.DefaultApi20;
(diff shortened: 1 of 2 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi20.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
kullfar: remove this templates pls

scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi20.java
```diff
@@ -0,0 +1,57 @@
...
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.github.scribejava.core.model.OAuthConfig;
+import com.github.scribejava.core.oauth.OAuth20Service;
+import com.github.scribejava.core.utils.OAuthEncoder;
+import com.github.scribejava.core.utils.Preconditions;
+
+/**
+ *
+ * @author jconner
+ */
```
kullfar: no need for author, git blame is better

kullfar:
Looks like template/example of usage. Can't imagine how it can be used in the production.
Put it in the wiki, here, please: https://github.com/scribejava/scribejava/wiki

PR #659, opened 2016-05-11
Fixed password grant type (tested with Salesforce)

Saal89:
Previous password grant type didn't work for my integration with salesforce. Adding those parameters resolved this.

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
@@ -172,15 +172,17 @@ public class OAuth20Service extends OAuthService {
 
         request.addParameter(OAuthConstants.GRANT_TYPE, OAuthConstants.PASSWORD);
 
-        request.addParameter(OAuthConstants.CLIENT_ID, config.getApiKey());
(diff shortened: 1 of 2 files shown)


scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
```diff
@@ -172,6 +172,8 @@ public final OAuth2AccessToken getAccessTokenPasswordGrant(String uname, String
 
         request.addParameter(OAuthConstants.GRANT_TYPE, OAuthConstants.PASSWORD);
 
+        request.addParameter(OAuthConstants.CLIENT_ID, config.getApiKey());
+        request.addParameter(OAuthConstants.CLIENT_SECRET, config.getApiSecret());
```
kullfar: according to https://tools.ietf.org/html/rfc6749#section-4.3.1
this data is not required to send.
Maybe it's better to make them optional?
Like this
```
if(config.hasClientCredentials()) {
...
}
```

Saal89: Thanks. I have updated my code

kullfar:
merged

PR #662, opened 2016-05-30
Add Misfit API

arrrrto:
Support for Misfit Wearables API

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/MisfitApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/MisfitApi.java
@@ -1,0 +1,56 @@
+package com.github.scribejava.apis;
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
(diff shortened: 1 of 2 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/MisfitApi.java
```diff
@@ -0,0 +1,56 @@
...
+            parameters.add(OAuthConstants.STATE, state);
+        }
+
+        return parameters.appendTo("https://api.misfitwearables.com/auth/dialog/authorize");
+    }
+
+    @Override
+    protected String getAuthorizationBaseUrl() {
+        throw new UnsupportedOperationException("use getAuthorizationUrl instead");
+    }
```
kullfar: Why not implement this method 
```
protected String getAuthorizationBaseUrl() {
return "https://api.misfitwearables.com/auth/dialog/authorize";
}
```
and remove getAuthorizationUrl here. It doesn't have any difference with overrided impl in DefaultApi20

PR #666, opened 2016-06-02
Add missing getRequestTokenAsync methods to OAuth10aService

bevilp:


kullfar:
it's not the best way to implement async version.
You should make some methods 'final' to not allow overriding them.
User can override 'prepareXXXRequest' method and get sync and async version of getXXX automagicaly.
here is the impl
https://github.com/scribejava/scribejava/commit/545f2d55288c31941a01b2b8daef27dad1206b9b
(already in master)

PR #667, opened 2016-06-10
Adds API provider for Salesforce

muenzpraeger:
This PR adds an API provider for Salesforce (OAuth 2.0).

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
@@ -6,7 +6,7 @@ import javax.net.ssl.SSLContext;
 import com.github.scribejava.apis.salesforce.SalesforceJsonTokenExtractor;
 
 /**
- * This class is an implementation of the Salesforce OAuth2 API. 
+ * This class is an implementation of the Salesforce OAuth2 API.
  */
 
 import com.github.scribejava.core.builder.api.DefaultApi20;
@@ -15,24 +15,24 @@ import com.github.scribejava.core.model.OAuth2AccessToken;
 import com.github.scribejava.core.model.Verb;
 
 public class SalesforceApi extends DefaultApi20 {
-	
-	// Constants
+
+    // Constants
     private static final String BASE_URL = "https://login.salesforce.com/services/oauth2";
     private static final String ACCESS_URL = BASE_URL + "/token?grant_type=authorization_code";
     private static final String AUTHORIZE_URL = BASE_URL + "/authorize";
 
     protected SalesforceApi() {
-    	SSLContext sc;
-		try {
-			String protocol = SSLContext.getDefault().getProtocol();
-			if (!protocol.equals("TLSv1.2")) { // This is important!
-				sc = SSLContext.getInstance("TLSv1.2");
-				sc.init(null, null, null);
-				HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());				
-			}
-		} catch (Exception e) {
-			throw new  IllegalStateException("Unexpected JVM without TLS 1.2 support", e);
-		} 
+        final SSLContext sc;
+        try {
+            final String protocol = SSLContext.getDefault().getProtocol();
+            if (!"TLSv1.2".equals(protocol)) { // This is important!
+                sc = SSLContext.getInstance("TLSv1.2");
+                sc.init(null, null, null);
+                HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
+            }
+        } catch (Exception e) {
+            throw new IllegalStateException("Unexpected JVM without TLS 1.2 support", e);
+        }
     }
 
     private static class InstanceHolder {
@@ -42,22 +42,22 @@ public class SalesforceApi extends DefaultApi20 {
     public static SalesforceApi instance() {
         return InstanceHolder.INSTANCE;
     }
-    
+
     @Override
-    public Verb getAccessTokenVerb(){
-          return Verb.POST;
+    public Verb getAccessTokenVerb() {
+        return Verb.POST;
     }
 
-	@Override
-	public String getAccessTokenEndpoint() {
-		return ACCESS_URL;
-	}
+    @Override
+    public String getAccessTokenEndpoint() {
+        return ACCESS_URL;
+    }
 
-	@Override
-	protected String getAuthorizationBaseUrl() {
-		return AUTHORIZE_URL;
-	}
-	
+    @Override
+    protected String getAuthorizationBaseUrl() {
+        return AUTHORIZE_URL;
+    }
+
     @Override
     public TokenExtractor<OAuth2AccessToken> getAccessTokenExtractor() {
         return SalesforceJsonTokenExtractor.instance();
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
@@ -28,7 +28,7 @@ public class SalesforceApi extends DefaultApi20 {
     	SSLContext sc;
 		try {
 			String protocol = SSLContext.getDefault().getProtocol();
-			if (!protocol.equals("TLSv1.2") || !protocol.equals("TLSv1.1")) { // This is important!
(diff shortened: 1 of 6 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.github.scribejava.core.extractors.TokenExtractor;
+import com.github.scribejava.core.model.OAuth2AccessToken;
+import com.github.scribejava.core.model.Verb;
+
+public class SalesforceApi extends DefaultApi20 {
+	
+	// Constants
```
kullfar: do not use tabs please. just whitspaces

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
```diff
@@ -0,0 +1,72 @@
...
+			if (!protocol.equals("TLSv1.2") || !protocol.equals("TLSv1.1")) { // This is important!
+				sc = SSLContext.getInstance("TLSv1.2");
+				sc.init(null, null, null);
+				HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());				
+			}
+		} catch (NoSuchAlgorithmException e) {
+			e.printStackTrace();
+		} 
+		catch (KeyManagementException e) {
+			e.printStackTrace();
```
kullfar: doesn't look like a good code.
We should notify users and let them know about this exceptions, aren't we?

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
```diff
@@ -0,0 +1,72 @@
...
+	private static final String AUTHORIZE_URL = BASE_URL + "/authorize";
+
+    protected SalesforceApi() {
+    	SSLContext sc;
+		try {
+			String protocol = SSLContext.getDefault().getProtocol();
+			if (!protocol.equals("TLSv1.2") || !protocol.equals("TLSv1.1")) { // This is important!
+				sc = SSLContext.getInstance("TLSv1.2");
+				sc.init(null, null, null);
+				HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
```
kullfar: Will it work with acyn http clients, like ning or AsyncHttpClient?

muenzpraeger: Good question, will check.

Main reason for this "not that good" implementation is, that Scribe itself doesn't enforce or allow via  configuration on outgoing HTTPS connections the usage of TLS. So if a server admin hasn't changed the JVM defaults it'll use SSLv3. And that's bad.

Most providers disable SSLv3 nowadays and switch to a higher encryption layer. The alternative tp this specific implementation would be to allow the configuration of the to be used TLS layer via the scribe-core package.

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
muenzpraeger: Will change

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
muenzpraeger: I agree. It doesn't look good. But the error itself is literally less than an edge case as TLS is a common implementation across all JVMs.

kullfar: Let them be thrown away then.
Don't catch them. Or throw new IllegalStateException("Unexpected JVM without xxx support", e);
It'll look better than printStackTrace and no info upper on the stack

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
muenzpraeger: Changed

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
muenzpraeger: Added new Throw

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
muenzpraeger: Added usage example with Async implementation

kullfar:
There are 303 errors reported by Checkstyle 6.19 with checkstyle.xml ruleset.

scribejava-apis/src/main/java/com/github/scribejava/apis/SalesforceApi.java
kullfar: they are still here
I think they are the main cause for checkstyle violations

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/SalesforceAsyncExample.java
```diff
@@ -0,0 +1,123 @@
...
+                .setRequestTimeout(10_000)
+                .setAllowPoolingConnections(false)
+                .setPooledConnectionIdleTimeout(1_000)
+                .setReadTimeout(10_000)
+                .build();
+
+        final OAuth20Service service = new ServiceBuilder()
+                .apiKey(clientId)
+                .apiSecret(clientSecret)
+                .asyncHttpClientConfig(clientConfig)
```
kullfar: btw, there is no such method anymore in master
it becomes asyncNingHttpClientConfig

muenzpraeger: I merged the current master to reflect the mentioned change in the fork.

kullfar:
Hi. I have reworked your pull request.
And created new one (fixuped together all your commits + some amends from me).
The main difference is - I don't like the way we automagically in background change SSLContext configuration.
I added only check to the SalesforceApi constructor https://github.com/scribejava/scribejava/pull/670/files#diff-4088d4bc2d35c02381574e4cdf0ad503R24
So user will get this stacktrace once he will invoke method SalesforceApi::instance. Like this
```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at com.github.scribejava.apis.SalesforceApi.instance(SalesforceApi.java:44)
	at com.github.scribejava.apis.examples.SalesforceAsyncExample.main(SalesforceAsyncExample.java:46)
Caused by: java.lang.IllegalStateException: Salesforce API required to use TLSv1.1 or upper. Enabled it by invoking method initTLSv11orUpper or somehow else
	at com.github.scribejava.apis.SalesforceApi.<init>(SalesforceApi.java:29)
	at com.github.scribejava.apis.SalesforceApi$InstanceHolder.<clinit>(SalesforceApi.java:40)
	... 2 more
```

And I added public static method to the SalesforceApi - initTLSv11orUpper https://github.com/scribejava/scribejava/pull/670/files#diff-4088d4bc2d35c02381574e4cdf0ad503R86
It will enable TLSv1.2 or TLSv1.1 support in case it's not already enabled and is supported by jvm at all.

I think, it's the best solution. Let the administrator or developer explicitly enable TLS >= v1.0 support in the way he prefer. Maybe diff params to init(null, null, null) method.

And in case of simple user with default jre7 he need just add this line of code 
```
SalesforceApi.initTLSv11orUpper();
```
Like I did in the example https://github.com/scribejava/scribejava/pull/670/files#diff-8c0011846b99499ea41dc09d9bc800e5R44
Without this line user will get Exception stack trace like I have showed above.

And user with JRE8 will not do anything. JRE8 have TLS1.2 enabled by default.

Look at this PR
https://github.com/scribejava/scribejava/pull/670
It's working with credentials you have provided.
If it looks ok from your point of view, I'll merge it to the master.

muenzpraeger:
I'm good with that. Basically it's your API. ;-)

kullfar:
Ooookey ;-)

kullfar:
Thanks for the good job

kullfar:
Oops. I merged this PR ;-(
fixing... revert + merge another one

Issue #668, opened 2016-06-14
ScribeJava shouldn't require Async libraries to be on the classpath


bdunn44:
The latest versions of Scribe have aimed to fix this, but there is still an issue with the OAuthConfig constructor which indirectly references ning/AHC classes and throws a compilation error. This should be fixable by adding a constructor to OAuthConfig without the Async client configuration.

kullfar:
Hm... And what is the user story to use this constructor outside the library?
Looks like OAuthConfig shoud be 'private' class instead

bdunn44:
I've developed a [Vaadin ](https://vaadin.com/home) add-on that uses this library. It allows users to tweak the OAuth configuration for the edge cases where the OOTB API configurations won't work. See [OAuthData.java](https://github.com/bdunn44/vaadin-oauthpopup/blob/master/oauthpopup/src/main/java/org/vaadin/addon/oauthpopup/OAuthData.java) (line 274). 

Essentially my add-on exposes it's own configuration object ([OAuthPopupConfig.java](https://github.com/bdunn44/vaadin-oauthpopup/blob/master/oauthpopup/src/main/java/org/vaadin/addon/oauthpopup/OAuthPopupConfig.java)), which produces a Scribe config object when needed (#asScribeConfig, line 89). This allows users to tweak details like the callback URL, etc.This isn't an issue with 2.4.0 but I'm looking to upgrade to the latest version without adding the ning/AHC dependencies to my project.

Is there another way that you would recommend I configure Scribe services other than using OAuthConfig? I'm not aware of another method, and it would seem quite limiting to remove that flexibility.

kullfar:
OAuthPopupConfig can have method like
```
createScribeServiceBuilder() {
new ServiceBuilder()
.apiKey(getApiKey())
.apiSecret(getApiSecret())
....
}
```
and then in  OAuthData.java it will be like this
```
	private OAuthService getService() {
		if (service == null) {
			service = config.createScribeServiceBuilder().build(isOAuth20()?api20:api10a);
		}
		return service;
}
```
isn't it good?

bdunn44:
Perfect! Thanks for the heads up on the ServiceBuilder. I think that will work for me.

Out of curiosity, what happened to the Grant Type configuration?

kullfar:
It doesn't needed while service building. So it was removed.
It's needed in the specific methods.

OAuthConstants.GRANT_TYPE=OAuthConstants.AUTHORIZATION_CODE
[...]

PR #669, opened 2016-06-14
Add constructor to avoid indirect ning/AHC class dependencies

bdunn44:
Fix for GitHub issue #668. Add a constructor without references to ning/AHC classes to remove indirect dependencies.


kullfar:
for history. closing, issue were solved, comments here: https://github.com/scribejava/scribejava/issues/668

PR #674, opened 2016-06-20
Fixed shouldExtractStandardHeader to not rely on the unguaranteed order

alexgyori:
The test assumes that the HashMap is iterated in a certain order.

The fix simply asserts that the required elements are contained in the header string in any order, insted of assuming a particular order.
It then asserts that only the 4 checked parts are in the string and are separated by comma, as the original test checks.
This way of checking makes the test robust to changes in the underlying JVM's implementation of HashMap.


kullfar:
Good, Thanks. Merged to the master

PR #676, opened 2016-06-21
fixed MapUtils test to not rely on HashMap's internal ordering

alexgyori:
MapUtils seems it should work for any instance of Map but this test is assuming a certain ordering from the HashMap; HashMap's spec does not guarantee any order.

Changed to LinkedHashMap, which guarantees the order; the pretty printing should work for any kind of map.


kullfar:
thx

PR #684, opened 2016-07-12
Fix #683 - Extract HTTP client logic into providers

### LINKED ISSUE #683: Adding some abstraction arround http client
I'd like to use scribejava with Okhttp where scribejava will use Okhttp.

I found something going into this direction: #484 and #485 

Is it something you can accept?

About implementation, we can imagine having a module like `scribejava-okhttp` (and some others) which is responsible to integrate scribejava and okhttp together. At least, this integration could be done in a 3rd party project.

juherr:
Proposition for #683

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,8 @@
     <modules>
         <module>scribejava-core</module>
         <module>scribejava-apis</module>
+        <module>scribejava-httpclient-ahc</module>
+        <module>scribejava-httpclient-ning</module>
     </modules>
 
     <licenses>
@@ -91,18 +93,6 @@
             <version>1.10</version>
             <scope>compile</scope>
             <optional>true</optional>
-        </dependency>
-        <dependency>
-            <groupId>com.ning</groupId>
-            <artifactId>async-http-client</artifactId>
-            <version>1.9.38</version>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.asynchttpclient</groupId>
-            <artifactId>async-http-client</artifactId>
-            <version>2.0.9</version>
-            <scope>provided</scope>
         </dependency>
     </dependencies>
     <build>
--- a/scribejava-core/src/main/java/com/github/scribejava/core/builder/ServiceBuilder.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/builder/ServiceBuilder.java
@@ -1,12 +1,14 @@
 package com.github.scribejava.core.builder;
 
 import com.github.scribejava.core.builder.api.BaseApi;
+import com.github.scribejava.core.model.HttpClient;
 import com.github.scribejava.core.model.OAuthConfig;
-import java.io.OutputStream;
 import com.github.scribejava.core.model.OAuthConstants;
 import com.github.scribejava.core.model.SignatureType;
 import com.github.scribejava.core.oauth.OAuthService;
 import com.github.scribejava.core.utils.Preconditions;
+
+import java.io.OutputStream;
 
 /**
  * Implementation of the Builder pattern, with a fluent interface that creates a {@link OAuthService}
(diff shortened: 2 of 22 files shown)


pom.xml
```diff
@@ -17,6 +17,8 @@
     <modules>
         <module>scribejava-core</module>
         <module>scribejava-apis</module>
+        <module>scribejava-ahc</module>
+        <module>scribejava-ning</module>
```
kullfar: Sometime ScribeJava will maybe have some new set of submodules.
Let this httpclient submodules pack will be named
scribejava-httpclient-ahc
and
scribejava-httpclient-ning

scribejava-ahc/src/main/java/com/github/scribejava/core/async/ahc/AhcHttpClient.java
```diff
@@ -0,0 +1,63 @@
+package com.github.scribejava.core.async.ahc;
```
kullfar: It's not in 'core' module now. And I think, "async" is not the valuable keyword already.
So as for me the package name should be just
com.github.scribejava.httpclient.ahc
to reflect maven arifact id

juherr:
@kullfar Your remarks are not a problem. Just say me when your review is done and I'll make updates.

scribejava-core/src/main/java/com/github/scribejava/core/async/HttpProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.github.scribejava.core.async;
+
+import com.github.scribejava.core.model.HttpClient;
+
+public interface HttpProvider {
```
kullfar: rename to HttpClientProvider pls
to reflect its client nature

scribejava-ahc/src/main/java/com/github/scribejava/core/async/ahc/AhcProvider.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class AhcProvider implements HttpProvider {
+
+    @Override
+    public HttpClient createClient(HttpClient.Config config) {
+        if (!(config instanceof AhcHttpClientConfig)) {
+            return null;
+        }
+        final AhcHttpClientConfig httpClientConfig = (AhcHttpClientConfig) config;
+        return new AhcHttpClient(httpClientConfig);
```
kullfar: just a bit more readable
```
        if (config instanceof AhcHttpClientConfig) {
            return new AhcHttpClient((AhcHttpClientConfig) config);
        }
        return null;
```
IMHO

scribejava-ahc/src/main/resources/META-INF/services/com.github.scribejava.core.async.HttpProvider
```diff
@@ -0,0 +1 @@
+com.github.scribejava.core.async.ahc.AhcProvider
```
kullfar: "No newline at end of file."
It's common to add newline at all files, let it be so, thx

scribejava-core/src/main/java/com/github/scribejava/core/builder/ServiceBuilder.java
```diff
@@ -28,11 +30,7 @@
     private Integer readTimeout;
 
     //async version only
```
kullfar: not relevant comment I thinks
smth like this
//not-default httpclient only

kullfar:
Great job!
Looks like the ideal, right what I had in mind, thx a lot!
I left some minor comments.

juherr:
Updated according to your remarks.

kullfar:
I tested the branch, all is awesome, merged with some small additions https://github.com/scribejava/scribejava/commit/8c2f85bcc269e85c9609ad78a65bf5ebcc7d4ebe

kullfar:
Thx again

PR #686, opened 2016-07-18
Introduce a integrated way to obtain code and manage token storage

juherr:


kullfar:
Looks like as unnecessary complications.
What are the needs in the CodeProvider? How are you planing to use it?
As for FileTokenStorage, imho it's out-of-scope for OAuth client library

scribejava-core/src/main/java/com/github/scribejava/core/model/AbstractTokenStorage.java
```diff
@@ -0,0 +1,34 @@
...
+import com.github.scribejava.core.oauth.OAuth20Service;
+
+import java.io.IOException;
+
+public abstract class AbstractTokenStorage {
+
+    private final OAuth20Service service;
+
+    protected AbstractTokenStorage(OAuth20Service service) {
+        this.service = service;
```
kullfar: it should check that service has codeProvider in it

scribejava-core/src/main/java/com/github/scribejava/core/model/AbstractTokenStorage.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public OAuth2AccessToken getOrGenerateToken() throws IOException {
+        OAuth2AccessToken token = loadToken();
+        if (token == null) {
+            token = service.getAccessToken();
```
kullfar: otherwise it will fall here.
It's better to fall as earlier as possible

juherr:
> What are the needs in the CodeProvider? How are you planing to use it?

I need it for the TokenStorage

> As for FileTokenStorage, imho it's out-of-scope for OAuth client library

Ok for `FileTokenStorage` which is more a sample (I can move it into api/test) but what about `AbstractTokenStorage`?

scribejava-core/src/main/java/com/github/scribejava/core/model/AbstractTokenStorage.java
juherr: Ok, good idea!

kullfar:
>>What are the needs in the CodeProvider? How are you planing to use it?

>I need it for the TokenStorage

You can integrate it in the TokenStorage:
```
public abstract class AbstractTokenStorage {
    private CodeProvider codeProvider;
...
    public OAuth2AccessToken getOrGenerateToken() throws IOException {
...
            token = service.getAccessToken(codeProvider.getCode(service));
...
    }
...
}
```

>>As for FileTokenStorage, imho it's out-of-scope for OAuth client library
>Ok for FileTokenStorage which is more a sample (I can move it into api/test) but what about AbstractTokenStorage?

And so it can be implemented not in the core of ScribeJava but in the code using ScribeJava.
It still looks like a code not for ScribeJava core but more specific for concrete application, not for OAuth library.

juherr:
> You can integrate it in the TokenStorage:

:+1: 

> It still looks like a code not for ScribeJava core but more specific for concrete application, not for OAuth library.

I think it is debatable. It is what I did for my own needs and I think token storage is something important when you deal with OAuth. I'm still looking for best practices for that and I'd love that libraries help me by providing some clues.

Feel free to close the pull-request if you think scribejava is not the good place for that :)

kullfar:
Thanks anyway!
I'm closing this PR for now.
ps. it can be leaved here https://github.com/scribejava/scribejava/wiki
as any other useful examples

PR #696, opened 2016-08-22
Manage error response for token request

juherr:
Sometimes the request is bad and, as a user, I'd like to be able to check what is the problem.

Currently, the exception is:
```
Caused by: com.github.scribejava.core.exceptions.OAuthException: Response body is incorrect. Can't extract a '"access_token"\s*:\s*"(\S*?)"' from this: '{"error_description":"unknown, invalid, or expired refresh token","error":"invalid_grant"}
'
	at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.extractParameter(OAuth2AccessTokenJsonExtractor.java:64)
	at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.extract(OAuth2AccessTokenJsonExtractor.java:37)
	at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.extract(OAuth2AccessTokenJsonExtractor.java:12)
	at com.github.scribejava.core.oauth.OAuth20Service.sendAccessTokenRequestSync(OAuth20Service.java:37)
	at com.github.scribejava.core.oauth.OAuth20Service.refreshAccessToken(OAuth20Service.java:93)
```

The pull-request replaces it by:
```
Caused by: com.github.scribejava.core.exceptions.OAuthRequestException: Get Access token error
	at com.github.scribejava.core.oauth.OAuth20Service.sendAccessTokenRequestSync(OAuth20Service.java:47)
	at com.github.scribejava.core.oauth.OAuth20Service.refreshAccessToken(OAuth20Service.java:108)
```

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/exceptions/OAuthRequestException.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/exceptions/OAuthRequestException.java
@@ -1,0 +1,17 @@
+package com.github.scribejava.core.exceptions;
+
+import com.github.scribejava.core.model.OAuth2ErrorResponse;
+
+public class OAuthRequestException extends OAuthException {
+
+    private final OAuth2ErrorResponse error;
+
+    public OAuthRequestException(String message, OAuth2ErrorResponse error) {
+        super(message);
+        this.error = error;
+    }
+
+    public OAuth2ErrorResponse getError() {
+        return error;
+    }
+}
--- a/scribejava-core/src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java
@@ -1,9 +1,13 @@
 package com.github.scribejava.core.extractors;
 
+import java.io.IOException;
+import java.net.URI;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import com.github.scribejava.core.exceptions.OAuthException;
 import com.github.scribejava.core.model.OAuth2AccessToken;
+import com.github.scribejava.core.model.OAuth2ErrorResponse;
+import com.github.scribejava.core.model.Response;
 import com.github.scribejava.core.utils.Preconditions;
 
 /**
@@ -16,6 +20,9 @@ public class OAuth2AccessTokenJsonExtractor implements TokenExtractor<OAuth2Acce
     private static final String EXPIRES_IN_REGEX = "\"expires_in\"\\s*:\\s*\"?(\\d*?)\"?\\D";
     private static final String REFRESH_TOKEN_REGEX = "\"refresh_token\"\\s*:\\s*\"(\\S*?)\"";
     private static final String SCOPE_REGEX = "\"scope\"\\s*:\\s*\"(\\S*?)\"";
+    private static final String ERROR_REGEX = "\"error\"\\s*:\\s*\"(\\S*?)\"";
+    private static final String ERROR_DESCRIPTION_REGEX = "\"error_description\"\\s*:\\s*\"([^\"]*?)\"";
+    private static final String ERROR_URI_REGEX = "\"error_uri\"\\s*:\\s*\"(\\S*?)\"";
 
     protected OAuth2AccessTokenJsonExtractor() {
     }
@@ -30,10 +37,34 @@ public class OAuth2AccessTokenJsonExtractor implements TokenExtractor<OAuth2Acce
     }
 
     @Override
-    public OAuth2AccessToken extract(String response) {
-        Preconditions.checkEmptyString(response,
+    public OAuth2AccessToken extract(Response response) throws IOException {
+        final String body = response.getBody();
+        Preconditions.checkEmptyString(body,
                 "Response body is incorrect. Can't extract a token from an empty string");
 
+        if (response.getCode() != 200) {
+            generateError(response.getBody());
+        }
+        return createToken(body);
+    }
+
+    // Related documentation: https://tools.ietf.org/html/rfc6749#section-5.2
+    private static void generateError(String response) {
+        final String errorInString = extractParameter(response, ERROR_REGEX, true);
+        final String errorDescription = extractParameter(response, ERROR_DESCRIPTION_REGEX, false);
+        final String errorUriInString = extractParameter(response, ERROR_URI_REGEX, false);
+        URI errorUri;
+        try {
+            errorUri = errorUriInString == null ? null : URI.create(errorUriInString);
+        } catch (IllegalArgumentException iae) {
+            errorUri = null;
+        }
+
+        throw new OAuth2ErrorResponse(OAuth2ErrorResponse.OAuthError.valueOf(errorInString), errorDescription,
+                errorUri, response);
+    }
+
+    private OAuth2AccessToken createToken(String response) {
         final String accessToken = extractParameter(response, ACCESS_TOKEN_REGEX, true);
         final String tokenType = extractParameter(response, TOKEN_TYPE_REGEX, false);
         final String expiresInString = extractParameter(response, EXPIRES_IN_REGEX, false);
--- a/scribejava-core/src/main/java/com/github/scribejava/core/extractors/TokenExtractor.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/extractors/TokenExtractor.java
@@ -1,6 +1,10 @@
 package com.github.scribejava.core.extractors;
 
+import com.github.scribejava.core.exceptions.OAuthException;
+import com.github.scribejava.core.model.Response;
 import com.github.scribejava.core.model.Token;
+
+import java.io.IOException;
 
 /**
  * Simple command object that extracts a concrete {@link Token} from a String
@@ -14,5 +18,5 @@ public interface TokenExtractor<T extends Token> {
      * @param response the contents of the response
      * @return OAuth access token
      */
-    T extract(String response);
+    T extract(Response response) throws IOException, OAuthException;
 }
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
@@ -1,0 +1,51 @@
+package com.github.scribejava.core.model;
+
+import com.github.scribejava.core.exceptions.OAuthException;
+
+import java.net.URI;
+
+/**
+ * Representing <a href="https://tools.ietf.org/html/rfc6749#section-5.2">"5.2. Error Response"</a>
+ */
+public class OAuth2ErrorResponse extends OAuthException {
+
+    public enum OAuthError {
+        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope
+    }
+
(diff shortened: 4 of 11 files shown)


kullfar:
Exception message "Response body is incorrect. Can't extract a '"access_token"\s*:\s*"(\S*?)"' from this: '{"error_description":"unknown, invalid, or expired refresh token","error":"invalid_grant"}"

was more informative in logs ;-)

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
```diff
@@ -0,0 +1,36 @@
+package com.github.scribejava.core.model;
+
+import java.net.URI;
+
+// https://tools.ietf.org/html/rfc6749#section-5.2
```
kullfar: convert to javadoc, pls?

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
```diff
@@ -0,0 +1,36 @@
...
+public class OAuth2ErrorResponse {
```
kullfar: Do we need this class and OAuthRequestException simultaneously?
Looks like we can have just one class

OAuth2ErrorResponse extends OAuthException
representing "5.2.  Error Response" from https://tools.ietf.org/html/rfc6749#section-5.2

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
```diff
@@ -34,7 +38,18 @@ public OAuth20Service(DefaultApi20 api, OAuthConfig config) {
...
-        return api.getAccessTokenExtractor().extract(request.send().getBody());
+        final Response response = request.send();
+        final TokenExtractor<OAuth2AccessToken> extractor = api.getAccessTokenExtractor();
+        if (response.getCode() != 200) {
+            if (extractor instanceof OAuth2AccessTokenJsonExtractor) {
+                final OAuth2AccessTokenJsonExtractor jsonExtractor = (OAuth2AccessTokenJsonExtractor) extractor;
+                throw new OAuthRequestException("Get Access token error",
+                        jsonExtractor.extractError(response.getBody()));
+            } else {
+                throw new OAuthException("Get Access token error: " + response.getBody());
```
kullfar: IMO it's better to hide all this logic in the single extractor method

like this
```
public interface TokenExtractor<T extends Token, E extends OAuthException> {

    /**
     * Extracts the concrete type of token from the contents of an Http Response
     *
     * @param response the contents of the response
     * @return OAuth access token
     */
    T extract(String response) throws E;
}

public class OAuth2AccessTokenJsonExtractor implements TokenExtractor<OAuth2AccessToken, OAuth2ErrorResponse> {
...
```
?

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public enum OAuthError {
+        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope
+    }
+
+    private final OAuthError error;
+    private final String errorDescription;
+    private final URI errorUri;
```
kullfar: + non-documented
private String rawResponse; to hold this data
'{"error_description":"unknown, invalid, or expired refresh token","error":"invalid_grant"}"

kullfar:
Looks good. It's a correct direction of improving the library. Thx.
Actually it was one of the next my planned improvements to the library ;-)

juherr:
Thanks for the review.

It was more a draft and I wanted to confirm the idea before spending too much time on it.

I will update the pull-request with your remarks :)

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
juherr: I think the important part here is the http return code.

In the `extract` method you don't have this information. So, the only solution is trying to parse the response as a valid one, then trying to parse it as an error one. 
I'm not feeling it good.

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
juherr: Ok, I will add it.

juherr: Do you prefer a parse method somewhere which is parsing the json, or do you prefer to have it in the constructor?

`OAuth2ErrorResponse(String rawResponse, OAuthError error, String errorDescription, URI errorUri)` vs `OAuth2ErrorResponse(String rawResponse)`

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
juherr: Yes :+1:

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
kullfar: all usages of this method (except tests) are the same
extract(response.getBody())
we can make
1.extract(response)
or
2.extract(response.getCode(), response.getBody())

And let the extractor decide how to interpret each response code

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
kullfar: OAuth2ErrorResponse(OAuthError error, String errorDescription, URI errorUri, String rawResponse)

and the rawResponse can be null in some cases

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
juherr: Ok, go for 1. :+1:

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
kullfar: Parsing algorithms can be different and the OAuth2ErrorResponse object will remain the same. or even just constructed (in tests) as usual class

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
kullfar: 1. - is better for production code, but harder for using in tests ;-)

juherr: I prefer 1. because it allows the parser to use extra information from the response.

But `response.getBody()` is throwing `IOException`.

1. `TokenExtractor.extract(response) throw OAuthException, IOException`
2. `TokenExtractor.extract(response.getCode(), response.getBody()) throw OAuthException`

You choose.

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
juherr: What should I take as exception message? `errorDescription`?

scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
kullfar: Agreed, 1. is better

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
kullfar: I would take them all.
like this
"{'error'='"+error+"', 'errorDescription'='"+errorDescription+"', ....

kullfar: more info - easier to understand the problem.
And it will not add overhead in normal flow

juherr:
Ping @kullfar

kullfar:
Sorry, I missed your new commit. I'll try to look it today

scribejava-core/src/main/java/com/github/scribejava/core/exceptions/OAuthRequestException.java
```diff
@@ -0,0 +1,17 @@
+package com.github.scribejava.core.exceptions;
+
+import com.github.scribejava.core.model.OAuth2ErrorResponse;
+
+public class OAuthRequestException extends OAuthException {
```
kullfar: seems like is redundant for now

scribejava-core/src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java
```diff
@@ -30,10 +37,34 @@ public static OAuth2AccessTokenJsonExtractor instance() {
...
+        Preconditions.checkEmptyString(body,
                 "Response body is incorrect. Can't extract a token from an empty string");
 
+        if (response.getCode() != 200) {
+            generateError(response.getBody());
+        }
+        return createToken(body);
+    }
+
+    // Related documentation: https://tools.ietf.org/html/rfc6749#section-5.2
```
kullfar: turn to javadoc, pls.
Why not?

scribejava-core/src/main/java/com/github/scribejava/core/extractors/TokenExtractor.java
```diff
@@ -14,5 +18,5 @@
      * @param response the contents of the response
```
kullfar: change javadoc as well.
it's not the contents anymore, but the whole reponse

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2ErrorResponse.java
```diff
@@ -0,0 +1,51 @@
+package com.github.scribejava.core.model;
+
+import com.github.scribejava.core.exceptions.OAuthException;
+
+import java.net.URI;
+
+/**
+ * Representing <a href="https://tools.ietf.org/html/rfc6749#section-5.2">"5.2. Error Response"</a>
+ */
+public class OAuth2ErrorResponse extends OAuthException {
```
kullfar: There are at least 2  Error Response in OAuth2 
second one is https://tools.ietf.org/html/rfc6749#section-7.2

I think it's better to name this one as OAuth2AccessTokenErrorResponse

kullfar:
PR is almost ok. I Have commented some moments.
+ add line to the changelog

kullfar:
I have merged to the master. All my comments are minor, I will fix them myself now in the next commit.
Thanks for the awesome inprovemnts to the ScribeJava!

juherr:
👍

kullfar:
that's it
https://github.com/scribejava/scribejava/commit/aa4e6188b07a814b006a03ab3cdc0cdddfb061aa

kullfar:
I'll release new version today

juherr:
Ok, thanks a lot!

PR #708, opened 2016-09-27
add support to get access token with additional params

jujusharp:
in some case, the server may need some additional params, so i can create request with some additional params to get the access token

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuth20Service.java
@@ -56,6 +56,19 @@ public class OAuth20Service extends OAuthService {
         return sendAccessTokenRequestSync(request);
     }
 
+    public final OAuth2AccessToken getAccessToken(String code, Map<String, String> additionalParams) throws IOException {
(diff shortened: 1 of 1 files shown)


kullfar:
Travis said "All checks have failed"

kullfar:
What are these cases?
If some standard flow, it will be better to add support for this param(s) in the core.
Some very nonstandard params, specific to the one concrete XXXAPI - it can have it's own XXXService.java with overrided createAccessTokenRequest method or hardcoded params in the access_token url String in XXXAPI.java (method getAccessTokenEndpoint)

I don't believe, that a lot of OAuth servers give opportunity to send additional params while getting access token ;-)

jujusharp:
In my case, The server need to double check which client view(client view means if the client is from web, mobile or pc) for the token, and this param is not only with the redirect url in the first step.

kullfar:
Can creating XXXService.java help you?
this type of classes intends for such custom logic.

jujusharp:
OK, I think the best choice is to create a custom service to satisfied the requirement.
Thanks again!

kullfar:
Welcome. You can provide this Custom Service and new API in a new PullRequest then ;-)

Issue #433, opened 2013-12-02
Scribe header case-sensitivity


zacharyatstreak:
HTTP headers are defined to be case-insensitive. The standard display I've seen is to have the first letter uppercase--can you lowercase the header name before looking for it in the Request object, so it returns the correct result when I use this standard format?  I've run into this twice now.

depsypher:
Looks like the headers being case sensitive breaks stackexchange currently.

StackExchange compresses responses whether you ask for them that way or not:
https://api.stackexchange.com/docs/compression

But the header in the response is "content-encoding" not ""Content-Encoding".

Calling `Response.getBody()` delegates to `parseBodyContents()`, which fails to find the gzip encoding here:
`if ("gzip".equals(getHeader("Content-Encoding"))) {`

So instead of calling getBody() I have to do this directly:
`StreamUtils.getGzipStreamContents(resp.getStream()))`

I would prefer that getBody() could detect gzip on stackexchange appropriately.
Thanks!

depsypher:
It appears that the reason Content-Encoding was coming in lower-cased in my case had to do with Google Appengine's implementation of HttpUrlConnection doing that normalization... In any case, it seems like a scenario that scribe should handle. I'll submit a PR.

kullfar:
Pull Request was merged to the master

PR #711, opened 2016-10-01
Ensure consistent Content-Encoding header name in Response

depsypher:
Support HttpURLConnection implementations such as the one google
appengine uses, which normalizes header names to lower case in the
implementation of getHeaderFields().

Relates to Issue #433


kullfar:
good job, thanks!

PR #717, opened 2016-11-02
Added API and tests for The Things Network.

jpmeijers:
Both for v1-staging and v2-preview.
You need to register your own client_id with TTN to use this library.
More info about The Things Network on:
http://slack.thethingsnetwork.org/
https://www.thethingsnetwork.org/forum/
https://www.thethingsnetwork.org/wiki/Home

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/TheThingsNetworkV1StagingApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/TheThingsNetworkV1StagingApi.java
@@ -1,0 +1,41 @@
+package com.github.scribejava.apis;
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor;
+import com.github.scribejava.core.extractors.TokenExtractor;
+import com.github.scribejava.core.model.OAuth2AccessToken;
+import com.github.scribejava.core.model.Verb;
+
+public class TheThingsNetworkV1StagingApi extends DefaultApi20 {
+
+    protected TheThingsNetworkV1StagingApi() {
+    }
+
+    private static class InstanceHolder {
+        private static final TheThingsNetworkV1StagingApi INSTANCE = new TheThingsNetworkV1StagingApi();
+    }
+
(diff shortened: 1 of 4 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/TheThingsNetworkV1StagingApi.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public static TheThingsNetworkV1StagingApi instance() {
+        return InstanceHolder.INSTANCE;
+    }
+
+    @Override
+    public Verb getAccessTokenVerb() {
+        return Verb.POST;
+    }
```
kullfar: it's the default standard Verb for OAuth2. No need to override this method

scribejava-apis/src/main/java/com/github/scribejava/apis/TheThingsNetworkV1StagingApi.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    protected String getAuthorizationBaseUrl() {
+        return "https://account.thethingsnetwork.org/users/authorize";
+    }
+
+    @Override
+    public TokenExtractor<OAuth2AccessToken> getAccessTokenExtractor() {
+        return OAuth2AccessTokenJsonExtractor.instance();
+    }
```
kullfar: no need to override that as well

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/TTNV1Example.java
```diff
@@ -0,0 +1,102 @@
...
+import com.github.scribejava.core.model.Response;
+import com.github.scribejava.core.model.Verb;
+import com.github.scribejava.core.oauth.OAuth20Service;
+
+import java.io.IOException;
+import java.net.URLDecoder;
+import java.util.Random;
+import java.util.Scanner;
+
+public final class TTNV1Example {
```
kullfar: should be named like Api class.
It's hard to find corresponding Example otherwise

kullfar:
I have merged it to the master. Thanks again!

PR #721, opened 2016-11-16
Update LiveApi.java

danilomagrini:
- change line 21 to Verb.POST to fix error "{"error":"invalid_request","error_description":"The provided request must be sent using the HTTP 'POST' method."}"
- update AuthorizationBaseURL


scribejava-apis/src/main/java/com/github/scribejava/apis/LiveApi.java
```diff
@@ -18,7 +18,7 @@ public static LiveApi instance() {
 
     @Override
     public Verb getAccessTokenVerb() {
-        return Verb.GET;
+        return Verb.POST;
     }
```
kullfar: Just drop the whole method.
It's the default Verb for DefaultApi20.

kullfar:
outdated

Issue #722, opened 2016-11-16
refreshAccessToken in OAuth20Service should be calling api.getRefreshTokenEndpoint 


vivin:
Currently it is calling `api.getAccessTokenEndpoint` while `refreshAccessTokenAsync` is calling `api.getRefreshTokenEndpoint`. By default `api.getRefreshTokenEndpoint` delegates to `api.getAccessTokenEndpoint`, but that can be overridden. However, doing so will not have the desired effect due to the fact that `refreshAccessTokenAsync` is calling the wrong method.

kullfar:
Yep. good point. thx

kullfar:
fixed here: https://github.com/scribejava/scribejava/pull/723

Neightly:
I found it when using in enterprice development, and I am wondering when it will be push to maven central repository.

kullfar:
In couple of hours

kullfar:
it's on the way. v3.3.0
Is on sonatype already. 15min - 2 hours to be synced to the maven central

kullfar:
done http://repo2.maven.apache.org/maven2/com/github/scribejava/scribejava-core/3.3.0/

PR #723, opened 2016-11-16
issue #722

vivin:
`OAuth20Service.refreshAccessToken` uses the correct URL by calling `api.getRefreshTokenEndpoint` instead of `api.getAccessTokenEndpoint`.

Very minor bugfix.


kullfar:
Thanks!

PR #724, opened 2016-11-17
Generify OAuthService with access token type

apomelov:
Generify OAuthService with access token type and pull up the abstract signRequest method. This small change helps to avoid code duplication when working with both types of social networks (OAuth 1.0 and 2.0).


kullfar:
nice, thanks!

PR #728, opened 2016-12-06
Using interface instead of impl in AhcHttpClient.

ggbt:
I'd like to use com.github.scribejava.httpclient.ahc.AhcHttpClient with my own implementation of AsyncHttpClient and not DefaultAsyncHttpClient.


kullfar:
thanks!

PR #729, opened 2016-12-08
Add support for specifying the desired lifetime of the access token.

sklakegg:
When using the _Implicit Grant Flow_, example is shown in https://dev.fitbit.com/docs/oauth2/ it would be beneficial to be able to set the desired lifetime of the access token.

I modified the code to support this.

expires_in

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/builder/ServiceBuilder.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/builder/ServiceBuilder.java
@@ -24,6 +24,7 @@ public class ServiceBuilder {
     private OutputStream debugStream;
     private String responseType = "code";
     private String userAgent;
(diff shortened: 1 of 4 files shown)


kullfar:
It's not OAuth2 thing, right? https://tools.ietf.org/html/rfc6749#section-4.2.1
Should be made as an optional additional param.

sklakegg:
True, it is an additional param. As far as I can tell I've added it to by null by default, unless something else is specified? Would it be possible to add this functionality?

kullfar:
Can't you just add one new XXXOAuthServiceImpl.java like other here https://github.com/scribejava/scribejava/tree/master/scribejava-apis/src/main/java/com/github/scribejava/apis/service

sklakegg:
Ah right, that worked wonders!
Thank you for a great library.

PR #749, opened 2017-03-17
Adding 'invalid_token' to `ErrorCode` enum

JesusSpiegel:
The "invalid_token" error value included in the OAuth 2.0 standard (https://tools.ietf.org/html/rfc6750#page-15 - 6.2.2.) is not covered in `OAuth2AccessTokenErrorResponse.ErrorCode` and it's causing an `IllegalArgumentException` due to scribejava is unable to parse the give error:

```java.lang.IllegalArgumentException: No enum constant com.github.scribejava.core.model.OAuth2AccessTokenErrorResponse.ErrorCode.invalid_token
at java.lang.Enum.valueOf(Enum.java:254)
at com.github.scribejava.core.model.OAuth2AccessTokenErrorResponse$ErrorCode.valueOf(OAuth2AccessTokenErrorResponse.java:14)
at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.generateError(OAuth2AccessTokenJsonExtractor.java:68)
at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.extract(OAuth2AccessTokenJsonExtractor.java:47)
at com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor.extract(OAuth2AccessTokenJsonExtractor.java:16)
at com.github.scribejava.core.oauth.OAuth20Service.sendAccessTokenRequestSync(OAuth20Service.java:37)

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
@@ -12,7 +12,8 @@ public class OAuth2AccessTokenErrorResponse extends OAuthException {
     private static final long serialVersionUID = 2309424849700276816L;
 
     public enum ErrorCode {
-        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope, invalid_token
+        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope,
+        invalid_token
     }
 
     private final ErrorCode errorCode;
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
@@ -12,7 +12,7 @@ public class OAuth2AccessTokenErrorResponse extends OAuthException {
     private static final long serialVersionUID = 2309424849700276816L;
 
     public enum ErrorCode {
-        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope
+        invalid_request, invalid_client, invalid_grant, unauthorized_client, unsupported_grant_type, invalid_scope, invalid_token
     }
 
     private final ErrorCode errorCode;

### CODE REVIEW

**Comment by kullfar** [comment]
Aha, you're right. it's an 'Bearer Token Usage - RFC 6750', addition to 'OAuth 2.0 Framework - RFC 6749'. Cool. I'll merge it to the master. But I'll add all additions. 'insufficient_scope; should be added as well to the enum.

**Comment by kullfar** [comment]
or not...
invalid_request, invalid_token and insufficient_scope from 'Bearer Token Usage' https://tools.ietf.org/html/rfc6750 are error codes for resource requests. And OAuth2AccessTokenErrorResponse.ErrorCode is error code for access_token request.
It even sounds strange. invalid_token error while requesting token ;-)
-----

   invalid_token
         The access token provided is expired, revoked, malformed, or
         invalid for other reasons.  The resource SHOULD respond with
         the HTTP 401 (Unauthorized) status code.  The client MAY
         request a new access token and retry the protected resource
         request.
------
(c) https://tools.ietf.org/html/rfc6750#section-6

When did you get such exception? Some API returns this code while requesting access_token?

**Comment by kullfar** [comment]
Thanks for info. Could you point to such OAuth provider docs or specs?
I have tried google OAuth. After token was revoked, I've tried to refresh token and got
```
Exception in thread "main" com.github.scribejava.core.model.OAuth2AccessTokenErrorResponse: {
 "error": "invalid_grant",
 "error_description": "Token has been expired or revoked."
}
```

deadlybulb:
This happens in EVE Online's OAuth implementation: https://eveonline-third-party-documentation.readthedocs.io/en/latest/sso/index.html

A refresh attempt on a revoked token for EVE Online will return invalid_token instead of invalid_grant.

I'm checking to see whether they use a standard library (in which case I can talk to the library maintainer), or whether they implemented this themselves (higher probability of improper interpretation of the spec).  I have to say I find the spec a bit grey on this subject, but it's likely you know it better than I.

echorebel:
With the spring security oauth package as referenced in #904 one can also use the same token endpoint with a refresh token to query a new access token as @deadlybulb already described with eve's/ccp implementation.
What I experienced so far with with spring that is that one gets a `invalid_grant` when the refresh token was invalidated/revoked and `invalid_token` if a refresh token is expired. A client might need to be able to handle both or either of them depending on the oauth implementation.
Afaik there is no expiry defined in any RFC for the refresh token expiry, but implementations might still go for it. Anyway I would still suggest to add the invalid_token as a possible error code.

kullfar:
understood. Will be fixed in https://github.com/scribejava/scribejava/pull/904

kullfar:
JFYI
Fix in the master
https://github.com/scribejava/scribejava/commit/030d76872fe371a84b5d05c6c3c33c34e8f611f1

Issue #753, opened 2017-03-28
Additional parameters when obtaining request token


vivin:
Please let me know if I am reading the spec wrong. Under section 6.1.1 of the OAuth 1.0a spec, it mentions that it is possible to provide additional parameters when obtaining a request token:

>6.1.1.  Consumer Obtains a Request Token

> To obtain a Request Token, the Consumer sends an HTTP request to the Service Provider's Request Token URL. The Service Provider documentation specifies the HTTP method for this request, and HTTP POST is RECOMMENDED. The request MUST be signed and contains the following parameters:
> 
> oauth_consumer_key: The Consumer Key.
> oauth_signature_method: The signature method the Consumer used to sign the request.
> oauth_signature: The signature as defined in Signing Requests.
> oauth_timestamp: As defined in Nonce and Timestamp.
> oauth_nonce: As defined in Nonce and Timestamp.
> oauth_version: OPTIONAL. If present, value MUST be 1.0 . Service Providers MUST assume the protocol version to be 1.0 if this parameter is not present. Service Providers' response to non-1.0 value is left undefined.
> oauth_callback: An absolute URL to which the Service Provider will redirect the User back when the Obtaining User Authorization step is completed. If the Consumer is unable to receive callbacks or a callback URL has been established via other means, the parameter value MUST be set to oob (case sensitive), to indicate an out-of-band configuration.
> **Additional parameters: Any additional parameters, as defined by the Service Provider.**

However, I notice that the `getRequestToken()` does not provide a way to specify additional parameters. It isn't possible to override as it is final, and I am unable to override `prepareRequestTokenRequest` (as I could in 3.x) in order to supply these parameters, since it not only creates the request within that method, but signs it there as well using the private `appendSignature` method.

kullfar:
It's better to use this final official RFC https://tools.ietf.org/html/rfc5849
interested section is
https://tools.ietf.org/html/rfc5849#section-2.1

as for your question... I see... So making methods addOAuthParams and  appendSignature protected will solve your issue?

vivin:
Yes, that would definitely help. You can make them final also, so that it will lessen the chances of someone messing up that core part by overriding those methods. 

I can submit a PR if need be.

vivin:
Another option is to override the `getRequestToken` and `getRequestTokenAsync` methods and provide ones that accept a `Map<String, String> additionalParameters` argument. What do you think?

vivin:
[...]

PR #755, opened 2017-03-29
issue #753

vivin:
Makes `addOAuthParams` and `appendSignature` `protected final`, so that they can be used from overridden `prepareRequestTokenRequest` methods in subclasses. Since it does not make sense to override these methods, they have been marked as `final` as well.


kullfar:
great. thanks, I will  merge it a bit later

vivin:
@kullfar Modified this PR so that the methods are just `protected` instead of `protected final`.

kullfar:
Thanks. I'm planning to check&merge it on the next week.
'final' are good for methods, like 'getAccessTokenAsync'. It doesn't have any OAuth-logic, only infrastructure logic, that shouldn't be customizable in any child class. And all other methods with 'business' logic, that can differ from one API provider to another should be overridable. That's my point

kullfar:
Merged, releasing to the central today

PR #763, opened 2017-04-24
Allow perms to be specified in Flickr Api (read, write, or delete)

rogerhu:


rogerhu:
These test failures don't look related?

scribejava-apis/src/main/java/com/github/scribejava/apis/FlickrApi.java
```diff
@@ -34,7 +44,13 @@ public String getAccessTokenEndpoint() {
...
     @Override
     public String getAuthorizationUrl(OAuth1RequestToken requestToken) {
-        return "https://www.flickr.com/services/oauth/authorize?oauth_token=" + requestToken.getToken();
+        String authUrl = String.format(AUTHORIZE_URL, requestToken.getToken());
+
+        if (perm != null) {
+            authUrl += "&perms=" + perm;
+        }
+
+        return authUrl;
```
kullfar: Isn't it the 'scope' in terms of OAuth? just another GET-param name.
It's better to add method getAuthorizationUrl with OAuthConfig as argument in DefaultApi10a (as it is in DefaultApi20).
And override it here, in FlickrApi.

Will it solve your issue?

kullfar: Seems, I'm wrong. disregard my comment, pls.
Your PullRequst looks ok, I will return to it a bit later

kullfar:
Yep. Test failures on travis is travis bug. don't worry.

scribejava-apis/src/main/java/com/github/scribejava/apis/FlickrApi.java
```diff
@@ -10,7 +10,16 @@
  */
 public class FlickrApi extends DefaultApi10a {
 
-    protected FlickrApi() {
+    private static final String AUTHORIZE_URL = "https://www.flickr.com/services/oauth/authorize?oauth_token=%s";
+
+    private String perm; /* read, write, or delete (delete includes read/write) */
```
kullfar: 1.maybe enum?
Is it closed set of value? do you have a link to the docs?
2.maybe final. So FlickrApi will stay immutable
3.is it required param? Maybe a batch of methods will be better?
public static FlickrApi read() {
public static FlickrApi write() {
public static FlickrApi delete() {
And user will choose which one to use.

scribejava-apis/src/main/java/com/github/scribejava/apis/FlickrApi.java
```diff
@@ -10,7 +10,16 @@
...
+
+    public FlickrApi() {
```
kullfar: It still shouldn't be public, right?

scribejava-apis/src/main/java/com/github/scribejava/apis/FlickrApi.java
```diff
@@ -21,6 +30,7 @@ public static FlickrApi instance() {
         return InstanceHolder.INSTANCE;
```
kullfar: do we need this instance at all. Depends on whether 'perms' is required.

rogerhu: perms is not required, things will default to read only if you don't ask for one

scribejava-apis/src/main/java/com/github/scribejava/apis/FlickrApi.java
rogerhu: See: https://www.flickr.com/services/api/auth.oauth.html

```
https://www.flickr.com/services/oauth/authorize

After getting the Request Token, your application needs to present the Flickr authorization page to the user, where they will be asked to give permission to your application to access their data. The authorization page will present the user with a list of permissions your application is requesting, as defined in the authentication flow settings.

Additionally, you can pass the optional perms= parameter, asking for read, write, or delete privileges. This parameter will override the setting defined in your application's authentication flow.
```

I think given the lack of an interface definition, defining read()/write()/delete() is confusing.  I created an enum as the constructor now.

PR #765, opened 2017-04-25
Add support for Android AsyncHttpClient

rogerhu:
Add support for Android AsyncHttpClient

This works for Android only, so not sure if you are open to accepting this PR.  Otherwise, it will go as a separate fork? :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
         <module>scribejava-httpclient-ahc</module>
         <module>scribejava-httpclient-ning</module>
         <module>scribejava-httpclient-okhttp</module>
+        <module>scribejava-httpclient-loopj</module>
     </modules>
 
     <licenses>
--- a/pom.xml
+++ b/pom.xml
@@ -104,7 +104,7 @@
                 <plugin>
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
-                    <version>3.3.0</version>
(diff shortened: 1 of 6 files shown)


kullfar:
Sorry, no. Seems, it's better to make fork.
Anyway, some comments:
1.There should be file
scribejava-apis/src/main/resources/META-INF/services/com.github.scribejava.core.httpclient.HttpClientProvider
```
com.github.scribejava.httpclient.loopj.HttpProvider
```
2.And class like this:
```
package com.github.scribejava.httpclient.loopj;

import com.github.scribejava.core.httpclient.HttpClientProvider;
import com.github.scribejava.core.httpclient.HttpClient;
import com.github.scribejava.core.httpclient.HttpClientConfig;

public class HttpProvider implements HttpClientProvider {

    @Override
    public HttpClient createClient(HttpClientConfig config) {
        if (config instanceof AsyncHttpClientConfig) {
            return new AsyncHttpClient();
        }
        return null;
    }
}
```

for this to work:
```
    private static HttpClient getClient(HttpClientConfig config) {
        for (HttpClientProvider provider : ServiceLoader.load(HttpClientProvider.class)) {
            final HttpClient client = provider.createClient(config);
            if (client != null) {
                return client;
            }
        }
        return null;
    }
``` 
https://github.com/scribejava/scribejava/blob/master/scribejava-core/src/main/java/com/github/scribejava/core/oauth/OAuthService.java#L43

PR #770, opened 2017-05-11
add netbase api

gjesse:


kullfar:
Hi, looks ok.
I need test user and app credentials to run the example before merging to the master.

gjesse:
I don't think i can provide those.. i don't work for netbase and added this as part of an integration i'm working on with them. I did run the test with our account credentials but I think we'd have to get netbase to provide some for you in order to run the test yourself.

kullfar:
Why do you think that netbase should provide them?
I don't see any problem if you register new test application and test user and send their credentials to me to run the example.

gjesse:
I don't have a way of registering another test user with netbase.

kullfar:
Pity, I have to close this Pull Request

gjesse:
well that's a bummer since i've tested it & it works, but it's no big deal.

PR #778, opened 2017-06-30
add Hub api and replace access patern

Faktorealchik:
added HubApi, not like others, because it`s non usual oauth, 
for reasons:
site/api not on original site, like
https://api.github.com/user
in hub api will be on your site, like 
mysite/hub/api, https://www.jetbrains.com/hub/ there you can read if you want

and other.. what I changed, it was src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java

reason: In my code, I can`t use space, but I was needed.
in code, you see: 

final OAuth20Service service = new ServiceBuilder(clientId)
               .apiSecret(clientSecret)
               .state(key)
               .scope("0-0-0-0-0 "+clientId) //you can add another scope
               .callback("http://example.com/callback")
               .build(HubApi.instance(HUB_SERVICE));

in "scope" I can`t add space because of other checks on hub,
 but if I will replace space on "%20" it will be %2520 because of code this framework(it replace % on %25), 
so, my opinion was to replace (//S*?) on (.*?), I think, it will not change anything.


kullfar:
not actual anymore? erroneously created?

Faktorealchik:
actual, but I haven`t passed test`s... 
mvn package compile passed with error`s

kullfar:
Closed, as agreed in the mail conversation. Info about using jetbrains should be added to wiki.

PR #780, opened 2017-07-11
Added Frappe.io OAuth2 API

revant:
OAuth 2 API for apps made with Frappe Framework https://frappe.io

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/FrappeApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/FrappeApi.java
@@ -4,7 +4,6 @@ import com.github.scribejava.apis.frappe.FrappeJsonTokenExtractor;
 import com.github.scribejava.core.builder.api.DefaultApi20;
 import com.github.scribejava.core.extractors.TokenExtractor;
 import com.github.scribejava.core.model.OAuth2AccessToken;
-import com.github.scribejava.core.model.Verb;
 
 public class FrappeApi extends DefaultApi20 {
 
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/FrappeApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/FrappeApi.java
@@ -1,5 +1,6 @@
 package com.github.scribejava.apis;
 
+import com.github.scribejava.apis.frappe.FrappeJsonTokenExtractor;
 import com.github.scribejava.core.builder.api.DefaultApi20;
 import com.github.scribejava.core.extractors.TokenExtractor;
 import com.github.scribejava.core.model.OAuth2AccessToken;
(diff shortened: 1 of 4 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/FrappeApi.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public void setServerURL(String serverURL) {
+        this.serverURL = serverURL;
+    }
+
+    @Override
+    public Verb getAccessTokenVerb() {
+        return Verb.POST;
+    }
```
kullfar: no need to override this method. It's POST by default

kullfar:
The main question is "how can I test this Pull Request". I need FrappeExample.java and app/user credentials to successfully run it.

revant:
Login with (Role - System Manager)
Testsite : https://scribejava.mntechique.com
username : scribejava@example.com
password : scribejava

I'll delete above credentials after testing.

Client Details Added in Example,

Login and change client details here 

https://scribejava.mntechnique.com/desk#Form/OAuth%20Client/37763ecdb0

In this case serverURL keeps changing as per the hosting domain name.

For others who wish to use their own frappe server url, I can think of extending FrappeApi to CustomFrappeApi and using it instead. I've added setter and getter for this.

Any thing better you can suggest?

e.g.

```
public class CustomFrappeAPI extends FrappeApi {
    protected CustomFrappeAPI(){
        setServerURL("http://custom.frappe.url.tld");
    }

    private static class InstanceHolder {
        private static final CustomFrappeAPI INSTANCE = new CustomFrappeAPI();
    }

    public static CustomFrappeAPI instance() {
        return CustomFrappeAPI.InstanceHolder.INSTANCE;
    }
}
```

References :

Frappe Framework is Open Source Web App Framework
Setup frappe server using frappe/bench - https://github.com/frappe/bench
Setup OAuth 2 and use REST API - http://frappe.github.io/frappe/user/en/guides/integration/
frappe framework https://github.com/frappe/frappe

kullfar:
great work, well done.
I have merged this Pull Request to the master with some additions
1.I created generic OpenId token and extractor for Google and Frappe.
2.As for custom serverUrl, it's a required param to Frappe. So, it's the required param for instance() method.
3.credentials are not needed anymore. I have tested Example, and removed credential from the sources.

kullfar:
I have fixed up all commits to the single one - https://github.com/scribejava/scribejava/commit/6dbf10e638cc74b22f2ad8c997e97b94d738213b
my additions - https://github.com/scribejava/scribejava/commit/f42812f4db9f49a98520c65de28fef50f380494d and https://github.com/scribejava/scribejava/commit/48d17c2f986af2e611c467d2f996c5d543fba859

PR #791, opened 2017-08-26
add ucoz api

evstropovv:
add [ucoz api](http://api.ucoz.net/ru/join/reg/pro), example


kullfar:
reviewed, corrected a bit, merged to the master.
Thanks! Awesome job

PR #824, opened 2018-02-10
Convert Tumblr api request urls to https

highthunder:
Tumblr seems to have changed their API to refuse requests that are not using HTTPS. 
These urls are listed [here](https://www.tumblr.com/docs/en/api/v2#auth).


kullfar:
Thanks for finding that.
Do you have test app and test user to run the Example, so I can test and merge this fix?

highthunder:
The included demo program demostrates this error and the fix, if you actually want to run the test, you'll need your own Tumblr API key + secret.
This is also duplicated by #825 .

kullfar:
I have no Tumblr API key + secret. (c) https://github.com/scribejava/scribejava/wiki/FAQ#when-will-you-support-insert_api_here-
And I still need them to run the example to merge this fix, sorry.

highthunder:
I've sent an email to the listed address containing all of the credentials needed to test this fix.

kullfar:
Thanks! I have merged this Pull Request to the master.

PR #828, opened 2018-02-20
Add OAuth2.0 for Yahoo APIs with example.

javatestcase:
Add OAuth2.0 for Yahoo APIs with example.


kullfar:
Hi,
Thanks for the Pull Request.

scribejava-apis/src/main/java/com/github/scribejava/apis/YahooApi20.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    protected String getAuthorizationBaseUrl() {
+        return "https://api.login.yahoo.com/oauth2/request_auth";
+    }
+
+    @Override
+    public TokenExtractor<OAuth2AccessToken> getAccessTokenExtractor() {
+        return OpenIdJsonTokenExtractor.instance();
```
kullfar: OpenIdJsonTokenExtractor adds openIdtoken from 'id_token' param.
I didn't see this param in the yahoo response.
I think there is no need in this OpenIdJsonTokenExtractor and default extractor can be used. So I guess this method can be safely dropped.

kullfar:
merged to the master

PR #837, opened 2018-03-07
Fixed missing support for scope for refresh_token grant_type.

tlxtellef:
https://tools.ietf.org/html/rfc6749#section-6


kullfar:
thanks. merged

PR #821, opened 2018-02-04
Add Wechat oauth2 support

zuoc:
Hello! 

I add Wechat(http://www.wechat.com/) support.

You can complete the test through the https://mp.weixin.qq.com/debug/cgi-bin/sandbox?t=sandbox/login.

Before that, you need to set up a server to receive WeChat‘s messages.

Here's the golang code I used to build the test server

```go
package main

import (
	"net/http"
	"log"
	"io"
)

func main() {
	http.HandleFunc("/message", func(w http.ResponseWriter, r *http.Request) {
		r.ParseForm()
		log.Println(r.Form)
		io.WriteString(w, r.Form.Get("echostr"))
	})

	http.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
		r.ParseForm()
		log.Println(r.Form)
	})

	log.Fatal(http.ListenAndServe(":8088", nil))
}

```

PR #739, opened 2017-02-10
Add WeChat API

xhusky:
Support for Wechat Open Platform

see [https://open.weixin.qq.com/cgi-bin/showdocument?action=dir_list&t=resource/res_list&verify=1&id=open1419316505&token=&lang=zh_CN](url)

PR #858, opened 2018-06-26
Modify the enum as an abstract type, allowing custom implementations.

zawn:
Modify the enum type to be an abstract type to allow custom implementations
For example, WeChat's OAuth2.0  client authentication parameter names are appid and secret


kullfar:
are such variants bad:
https://github.com/scribejava/scribejava/pull/821/files
https://github.com/scribejava/scribejava/pull/739
As far as I remember, they was ok with enum

zawn:
This implementation overrides the createAccessTokenRequest method of OAuth20Service and accomplishes similar functions 。
image
image
image

This is not necessary. All we have to do is add a custom client authentication method. This violates the original intention of the ClientAuthenticationType design.

zawn:
Of course, no matter how wechat is implemented, our focus is on https://tools.ietf.org/html/rfc6749#section-2.3.2

image
We should provide a possible way for users to implement

kullfar:
It does make sense. Thank for pointing to this. I need some time to think about this.

zawn:
This is only a minor change and will not affect other existing classes.
you can accept the pull request without additional burden

PR #860, opened 2018-07-06
modify the enum as an abstract type, allowing custom implementations。

zawn:
#858 Switch a branch


kullfar:
Eventualy, it's in the master. With some refactorings from me.
1.save enum for transparent upgrade, just mark it as deprecated for one version
2.make interface in the separate package, not an abstract class

PR #878, opened 2018-10-12
Keycloak api and keyclak example

JureZelic:
Added keycloak api and keycloak example. Realm can be added using setRealm methode of a KeyloackApi class. Everithing else is standard stuff.

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java
@@ -5,38 +5,47 @@ import com.github.scribejava.core.builder.api.DefaultApi20;
 import com.github.scribejava.core.extractors.TokenExtractor;
 import com.github.scribejava.core.model.OAuth2AccessToken;
 
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 public class KeycloakApi extends DefaultApi20 {
 
-    private String baseUrl = "http://localhost:8080";
-    private String realm = "master";
+    private final String baseUrlWithRealm;
 
-    public void setBaseUrl(String baseUrl) {
-        this.baseUrl = baseUrl + (baseUrl.endsWith("/")? "" : "/");
-    }
+    private static final ConcurrentMap<String, KeycloakApi> INSTANCES = new ConcurrentHashMap<>();
 
-    public void setRealm(String realm) {
-        this.realm = realm;
-    }
-
-    protected KeycloakApi() {
-    }
-
-    private static class InstanceHolder {
-        private static final KeycloakApi INSTANCE = new KeycloakApi();
+    protected KeycloakApi(String baseUrlWithRealm) {
+        this.baseUrlWithRealm = baseUrlWithRealm;
     }
 
     public static KeycloakApi instance() {
-        return KeycloakApi.InstanceHolder.INSTANCE;
+        return instance("http://localhost:8080/", "master");
+    }
+
+    public static KeycloakApi instance(String baseUrl, String realm) {
+        final String defaultBaseUrlWithRealm = composeBaseUrlWithRealm(baseUrl, realm);
+
+        //java8: switch to ConcurrentMap::computeIfAbsent
+        KeycloakApi api = INSTANCES.get(defaultBaseUrlWithRealm);
+        if (api == null) {
+            api = new KeycloakApi(defaultBaseUrlWithRealm);
+            INSTANCES.putIfAbsent(defaultBaseUrlWithRealm, api);
+        }
+        return api;
+    }
+
+    protected static String composeBaseUrlWithRealm(String baseUrl, String realm) {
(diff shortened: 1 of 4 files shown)


kullfar:
Hi,
Thanks for the Pull Request.
(c) https://github.com/scribejava/scribejava/wiki/FAQ#when-will-you-support-insert_api_here-

JureZelic:
Normally one would install keycloak server to protect it's application first, but I have created a test user on my test server.
1. First please change apiSecret (line 29 in KeycloakExample to "a8fe62c9-c3a1-4545-81ca-fda5df1c032b"
2. You must trust https://sicas.setcce.si server certificate (attached)
[sicassetccesi.crt.txt](https://github.com/scribejava/scribejava/files/2478302/sicassetccesi.crt.txt)

One more think. I just noticed I must rename Keyloack to Keycloak (I must be a bit dyslectic :-)). I will first do that.

JureZelic:
You can test now with credentials:
username: test
password: test123

scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.github.scribejava.apis.openid.OpenIdJsonTokenExtractor;
+import com.github.scribejava.core.builder.api.DefaultApi20;
+import com.github.scribejava.core.extractors.TokenExtractor;
+import com.github.scribejava.core.model.OAuth2AccessToken;
+
+public class KeycloakApi extends DefaultApi20 {
+
+    private String baseUrl = "http://localhost:8080";
+    private String realm = "master";
```
kullfar: I'd prefer API classes to be immutable.
And if we need some customisation, then I'd make method like this
```
    public static FacebookApi customVersion(String version) {
        return new FacebookApi(version);
    }
```
https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/main/java/com/github/scribejava/apis/FacebookApi.java#L39

I suggest such impl from a programmatic point of view
```
public class KeycloakApi extends DefaultApi20 {

    private final String baseUrlWithRealm;

    private static final ConcurrentMap<String, KeycloakApi> INSTANCES = new ConcurrentHashMap<>();

    protected KeycloakApi(String baseUrlWithRealm) {
        this.baseUrlWithRealm = baseUrlWithRealm;
    }

    public static KeycloakApi instance() {
        return instance("http://localhost:8080/", "master");
    }

    public static KeycloakApi instance(String baseUrl, String realm) {
        final String defaultBaseUrlWithRealm = composeBaseUrlWithRealm(baseUrl, realm);

        //java8: switch to ConcurrentMap::computeIfAbsent
        KeycloakApi api = INSTANCES.get(defaultBaseUrlWithRealm);
        if (api == null) {
            api = INSTANCES.putIfAbsent(defaultBaseUrlWithRealm, new KeycloakApi(defaultBaseUrlWithRealm));
        }
        return api;
    }

    protected static String composeBaseUrlWithRealm(String baseUrl, String realm) {
        return baseUrl + (baseUrl.endsWith("/") ? "" : "/") + "auth/realms/" + realm;
    }

    @Override
    public String getAccessTokenEndpoint() {
        return baseUrlWithRealm + "/protocol/openid-connect/token";
    }

    @Override
    protected String getAuthorizationBaseUrl() {
        return baseUrlWithRealm + "/protocol/openid-connect/auth";
    }
...
}
```
Is it ok from the business side?

kullfar: and usage:
```
        final OAuth20Service service = new ServiceBuilder(apiKey)
                .apiSecret(apiSecret)
                .scope("openid")
                .callback("http://najdi.si")
                .build(KeycloakApi.instance(BASE_URL, REALM));
```

JureZelic:
Great. I have corrected as You suggested (one minor change in KeycloakApi instance(String baseUrl, String realm) method) and tested.

scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static KeycloakApi instance(String baseUrl, String realm) {
+        final String defaultBaseUrlWithRealm = composeBaseUrlWithRealm(baseUrl, realm);
+
+        //java8: switch to ConcurrentMap::computeIfAbsent
+        KeycloakApi api = INSTANCES.get(defaultBaseUrlWithRealm);
+        if (api == null) {
+            api = new KeycloakApi(defaultBaseUrlWithRealm);
+            INSTANCES.putIfAbsent(defaultBaseUrlWithRealm, api);
```
kullfar: The error is here as well ;-)
Method putIfAbsent can ignore second param "api" and doesn't put in the map in case another thread will invoke this method in parallel, and method instance wll return another KeycloakApi instance.
will fix it in this manner:
```
alreadyCreatedApi = INSTANCES.putIfAbsent(defaultBaseUrlWithRealm, api);
    if(alreadyCreatedApi!=null) {
        return alreadyCreatedApi;
    }
```

PR #884, opened 2018-12-05
Fix and extend Multipart functionality of JDK Http Client

NTPape:
Hi,

I tried to use the Multipart functionality that was recently added to the `JDKHttpClient`. I specifically wanted to create a `multipart/form-data` request including files. But I found it had problems with both the headers as well as the body, but with this PR it works now.

First of all the Content-Type header was not specifically set by the `JDKHttpClient` for a Multipart request, so the method `prepareConnectionForBodyAndGetOutputStream` would always default to `application/x-www-form-urlencoded`. Moreover it was not possible to specify a Multipart subtype, which I also tried to fix with this PR.

The body was also wrong: the `getStartBoundary` didn't start with a CRLF, and the `getEndBoundary` was used after every BodyPart instead of only at the end.

I tried to keep it backward compatible by introducing a new method `initMultipartPayload` alongside the existing `initMultipartBoundary` in the `OAuthRequest`.

Thanks for looking into this! Please let me know if you have any feedback!

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
@@ -128,32 +128,61 @@ public class OAuthRequest {
     }
 
     /**
-     * Set boundary of multipart request
+     * Set subtype and boundary of multipart request
      *
-     * @param boundary can be any string
+     * @param subtype subtype of multipart content-type
+     * @param boundary can be any string between 1 and 70 characters long using alphanumeric characters and any of
+     *  "'" / "(" / ")" / "+" / "_" / "," / "-" / "." / "/" / ":" / "=" / "?" / " ", but must not end with white space
      */
-    public void initMultipartBoundary(String boundary) {
-        multipartPayload = new MultipartPayload(boundary == null
-                ? Long.toString(System.currentTimeMillis())
+    public void initMultipartPayload(String subtype, String boundary) {
+        multipartPayload = new MultipartPayload(subtype == null ? "mixed" : subtype, boundary == null
+                ? "---------------------------" + Long.toString(System.currentTimeMillis())
                 : boundary);
     }
 
     /**
-     * init boundary of multipart request with default boundary
+     * Init multipart request with given subtype and default boundary
+     *
+     * @param subtype subtype of multipart content-type
      */
-    public void initMultipartBoundary() {
-        initMultipartBoundary(null);
+    public void initMultipartPayload(String subtype) {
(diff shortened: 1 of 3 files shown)


kullfar:
Oh. We definitely need java11 and its HttpClient here ;-)
Ok, looking this PR to merge to the master

scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
```diff
@@ -128,32 +128,61 @@ public void addParameter(String key, String value) {
...
+     * @param subtype subtype of multipart content-type
+     * @param boundary can be any string between 1 and 70 characters long using alphanumeric characters and any of
+     *  "'" / "(" / ")" / "+" / "_" / "," / "-" / "." / "/" / ":" / "=" / "?" / " ", but must not end with white space
      */
-    public void initMultipartBoundary(String boundary) {
-        multipartPayload = new MultipartPayload(boundary == null
-                ? Long.toString(System.currentTimeMillis())
+    public void initMultipartPayload(String subtype, String boundary) {
+        multipartPayload = new MultipartPayload(subtype == null ? "mixed" : subtype, boundary == null
+                ? "---------------------------" + Long.toString(System.currentTimeMillis())
```
kullfar: what for is this string "---------------------------" here?

NTPape: I don't think it's technically necessary, but both Firefox and Chrome do it in some form or another for their multipart form-data requests (which is what I needed and tried to emulate and got working with this fix), so I thought it may be good practice. But I don't know why they do it exactly. I think it might very well be possible there's more to it than just a fixed number of dashes, too. So it can probably be removed without losing anything.

kullfar:
I have reworked multipart model, and its 'service' part in JDK HTTP client.
https://github.com/scribejava/scribejava/commit/fac5e70bee6d2bcd7865feb90cef3aeea648b91f
I have pushed changes to the master.
You can try and test it.
As example you can use this unit test
https://github.com/scribejava/scribejava/blob/fac5e70bee6d2bcd7865feb90cef3aeea648b91f/scribejava-core/src/test/java/com/github/scribejava/core/httpclient/jdk/JDKHttpClientTest.java


ps.Thanks again for fixing these problems, I have used them while reworking the code.

NTPape:
Hi @kullfar,
Thank you for your work!

Sorry that I only got around to testing this now.

Only thing that is mildly inconvenient is that I have to manually set the Content-disposition header for non-file parts to give them a name. But that's nothing major.

Apart from that it works great now, thanks again! :+1:

PR #897, opened 2019-03-07
Added an OpenIdTokenExtractor to MicrosoftAzureActiveDirectory20Api

jacobmims:
Since the AD 2.0 API is OpenId compliant, this allows users to extract OpenId information when the scope is granted. This is inline with the Google and StackExchange APIs.


kullfar:
Hi, thanks for the PullRequest!
I just need credentials of test application and test user to run the example and to test changes before merging them to the master.
Thanks.

jacobmims:
Any existing test case should work. This isn't fundamentally changing the behavior of the API implementation. It's only extracting the OpenId token that is given back when requesting the "openid" scope. If you still need a new app to test with, I can spin one up.

kullfar:
You are right, but I need to test it myself to guarantee good quality of this library.
I'll be grateful for a test app and test user to run the example.

PR #904, opened 2019-04-10
extend error codes with invalid_token

echorebel:
according to RFC6750 Section 6.2.2 error extension
see https://tools.ietf.org/html/rfc6750#section-6.2.2

makes devs live bit easier handling errors via spring-security-oauth, 
see https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/exceptions/OAuth2Exception.java

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuth2AccessTokenErrorResponse.java
@@ -15,6 +15,7 @@ public class OAuth2AccessTokenErrorResponse extends OAuthException {
         INVALID_REQUEST("invalid_request"),
         INVALID_CLIENT("invalid_client"),
         INVALID_GRANT("invalid_grant"),
+        INVALID_TOKEN("invalid_token"),
         UNAUTHORIZED_CLIENT("unauthorized_client"),
         UNSUPPORTED_GRANT_TYPE("unsupported_grant_type"),
         INVALID_SCOPE("invalid_scope"),


kullfar:
It looks like incorrect place for this error value.
Here was the same Pull Request here already. Look for the details here
https://github.com/scribejava/scribejava/pull/749

ScribeJava is not (yet) parse responses for API requests after AccessToken was received. only https://github.com/scribejava/scribejava/blob/master/scribejava-core/src/main/java/com/github/scribejava/core/model/Response.java is returned.
May be some time later it will, but not yet.

echorebel:
This is for refresh token usage on the endpoint to refresh the access token. When the refresh token is invalid the spring-security-oauth would respond with that error code. Parsing is already done in `OAuth2AccessTokenJsonExtractor`.
My situation now is that the exception error code is null just because it doesn't know the type. So we end up parsing again the raw response which could be avoided pretty simple.

kullfar:
I see... I have the idea. I'll move ErrorCode from the OAuth2AccessTokenErrorResponse. And it will represent whole OAuth Error Registry and it will be "used with the authorization code grant error response (Section 4.1.2.1), the implicit grant error response (Section 4.2.2.1), the token error response (Section 5.2), or the resource access error response (Section 7.2)"

kullfar:
Fix in the master
https://github.com/scribejava/scribejava/commit/030d76872fe371a84b5d05c6c3c33c34e8f611f1
thanks for the collaboration!

echorebel:
awesome thx!

PR #906, opened 2019-05-02
Use proper Authentication for keycloak

dborisenko:
I believe, `RequestBodyAuthenticationScheme` is the proper type of authentication, but not `HttpBasicAuthenticationScheme` one. At least, it doesn't give error anymore


dborisenko:
Otherwise it throws following exception:
```
 com.github.scribejava.core.model.OAuth2AccessTokenErrorResponse: {"error":"unauthorized_client","error_description":"INVALID_CREDENTIALS: Invalid client credentials"}
```

kullfar:
Hi, thanks for the Pull Request!
Could you provide me with test user credentials and test app client id and secret to run the example? (c) https://github.com/scribejava/scribejava/wiki/FAQ#when-will-you-support-insert_api_here-

kullfar:
outdated

PR #919, opened 2019-08-01
Add debug output to OAuth2Service

rbarbey:
I pulled up the `debugStream` field into `OAuthService` so that both `OAuth10aService` and `OAuth20Service` can provide debug output. I also added log messages to the `OAuth20Service` when different requestz are created and sent, both for the sync and async case.

To print `ParameterList` objects, I used the `asFormUrlEncodedString()` method because that was already there, knowing that the output format might not always be what you want.

Please let me know if there is anything you want me to change or add.

Thank you!


kullfar:
Hi! thanks for the PR. I have just merged it to the master with some small additions.

kullfar:
https://github.com/scribejava/scribejava/commit/32273dbcabe7a84c2324e4d5fbae1120e99579a0

PR #921, opened 2019-08-12
Added Xero API support

SidneyAllen:
- Removed old XeroApi20 which was for oAuth 1.0a but never submitted PR
- Added new Xero20Api which is for oAuth 2 urls and instance holder
- Added new XeroExample which demonstrates the scribe library.

You'll need to create a (free) Xero user account
https://www.xero.com/us/signup/api/

Then create an oAuth2 app at  https://developer.xero.com/myapps

You'll get an client id and client secret and set a redirect URI.  Once these 3 values are set in the XeroExample, you can run it to obtain an access token and access the "connections" endpoint a protected resource that returns and array of Xero tenant ids.  You'll set the xero tenant id in the header of any calls to Xero API protected resources.


kullfar:
Hi,
Thanks for the Pull Request.

Thanks in advance

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/XeroExample.java
```diff
@@ -0,0 +1,130 @@
...
+ * uncomment the rest of the example code to access
+ * the Xero Organisation and other protected resources
+ * 
+ * <dependency>
+        <groupId>com.googlecode.json-simple</groupId>
+        <artifactId>json-simple</artifactId>
+        <version>1.1.1</version>
+    </dependency>
+ * 
+ */
```
kullfar: There is the com.fasterxml.jackson library as the dependency in the ScribeJava, you can use it.

scribejava-apis/src/test/java/com/github/scribejava/apis/examples/XeroExample.java
```diff
@@ -0,0 +1,130 @@
...
+        // Now let's go and ask for a protected resource!
+        System.out.println("Got it! Lets see what we found...");
+        System.out.println();
+        System.out.println(response.getCode());
+        System.out.println(response.getBody());
+        */
+        System.out.println();
+        System.out.println("Thats it man! Go and build something awesome with ScribeJava! :)");
+    }
+}
```
kullfar: newline is needed

SidneyAllen:
I've updated the XeroExample to use Jackson to parse responses.  I will email you credentials and login for testing

kullfar:
I have merged this PR with some changes https://github.com/scribejava/scribejava/commit/177a977fa6e85757b1a5b4c243c17f532d661965

Thx for the Pull Request!

scribejava-apis/src/main/java/com/github/scribejava/apis/XeroApi20.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    protected String getAuthorizationBaseUrl() {
+        return "https://login.xero.com/identity/connect/authorize";
+    }
+
+    @Override
+    public ClientAuthentication getClientAuthentication() {
+        return RequestBodyAuthenticationScheme.instance();
+    }
```
kullfar: There was no need to Override this method. Default method (and the preferred one) is working with xero as well.

PR #936, opened 2019-10-15
feat: added polar oauth2 api and example

vidi42:


scribejava-apis/src/main/java/com/github/scribejava/apis/polar/PolarOAuth20Service.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class PolarOAuth20Service extends OAuth20Service {
+
+	public PolarOAuth20Service(DefaultApi20 api, String apiKey, String apiSecret, String callback, String defaultScope, String responseType,
+		OutputStream debugStream, String userAgent, HttpClientConfig httpClientConfig, HttpClient httpClient) {
+		super(api, apiKey, apiSecret, callback, defaultScope, responseType, debugStream, userAgent, httpClientConfig, httpClient);
+	}
+
+	@Override
+	protected OAuthRequest createAccessTokenRequest(AccessTokenRequestParams params) {
```
kullfar: This override is unneeded. It's the only diff is using "request.addBodyParameter" instead of "request.addParameter". But you didn't override getAccessTokenVerb in API. It's POST by default. So "request.addParameter" will invoke "request.addBodyParameter" inside.

ps.The second diff is questionable. You skipped setting OAuthConstants.SCOPE parameter. why?

kullfar:
Hi,
Thanks for the Pull Request.

Thanks in advance

scribejava-apis/src/main/java/com/github/scribejava/apis/polar/PolarOAuth20Service.java
vidi42: Yeah, the `request.addBodyParamete` was indeed not needed for POST requests. I've reverted that. However the `OAuthConstants.SCOPE` has to be left out for access token request otherwise the Polar API will respond with a 400, as described in their API https://www.polar.com/accesslink-api/#token-endpoint

kullfar:
Hi, I have merged this PR with small changes to the master.
One note - I always got 204 in Example. no data. I have even changed and verified the email for Polar account, no success (
But 204 is not a 403. I think it's a business issue, not an OAuth one.

kullfar:
Thanks again

PR #945, opened 2019-12-10
Close OkHttp response body

drei01:
Response body must be closed, see https://github.com/square/okhttp/issues/2311

### DIFF
--- a/scribejava-httpclient-okhttp/src/main/java/com/github/scribejava/httpclient/okhttp/OkHttpHttpClient.java
+++ b/scribejava-httpclient-okhttp/src/main/java/com/github/scribejava/httpclient/okhttp/OkHttpHttpClient.java
@@ -191,9 +191,10 @@ public class OkHttpHttpClient implements HttpClient {
             headersMap.put(headerName, headers.get(headerName));
         }
 
-        final ResponseBody body = okHttpResponse.body();
-        return new Response(okHttpResponse.code(), okHttpResponse.message(), headersMap,
-                body == null ? null : body.byteStream());
+        try (ResponseBody body = okHttpResponse.body()) {
+            return new Response(okHttpResponse.code(), okHttpResponse.message(), headersMap,
+                    body == null ? null : body.byteStream());
+        }
     }
 
 }


echorebel:
how is this change closing the body? can u explain?

scribejava-httpclient-okhttp/src/main/java/com/github/scribejava/httpclient/okhttp/OkHttpHttpClient.java
```diff
@@ -191,9 +191,10 @@ static Response convertResponse(okhttp3.Response okHttpResponse) {
...
         }
 
-        final ResponseBody body = okHttpResponse.body();
-        return new Response(okHttpResponse.code(), okHttpResponse.message(), headersMap,
-                body == null ? null : body.byteStream());
+        try (ResponseBody body = okHttpResponse.body()) {
+            return new Response(okHttpResponse.code(), okHttpResponse.message(), headersMap,
+                    body == null ? null : body.byteStream());
+        }
     }
```
echorebel: shouldn't there be a finally? i don't get how this closes the body.

drei01: This is a try-with-resources. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

echorebel: ok cool

kullfar: Hi, @drei01 , ResponseBody shouldn't be closed here in such way.
it is passed to the com.github.scribejava.core.model.Response.
and is used by it to read body stream later.
Developers should close com.github.scribejava.core.model.Response. And it will close the stream.
Response implements Closeable and have method
```
    @Override
    public void close() throws IOException {
        if (stream != null) {
            stream.close();
        }
    }
```

drei01: @kullfar thanks for the clarification. I'll close this PR.

kullfar:
But.... You're right, that only byteStream will be closed, not the ResponseBody. I will make some changes soon.

kullfar:
that's it
https://github.com/scribejava/scribejava/commit/b2458a8abe0cf8237c491986ba2832031bdcfd33

PR #952, opened 2020-03-03
pom.xml: specify Maven Central repository

simon04:
Otherwise dependencies cannot be resolved, unless they are cached locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,13 @@
         </developer>
     </developers>
 
+    <repositories>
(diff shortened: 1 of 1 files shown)


kullfar:
Hi, Simon!

Why? Isn't maven central is always in main maven config? I don't see the reason to duplicate this knowledge in pom.xml.

simon04:
I think you're right. I do no longer recall the exact problem from March. So let's just close this PR…

PR #958, opened 2020-04-15
initial support for device code flow

rebarbora-mckvak:
Tested on azure.

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/microsoftazureactivedirectory/BaseMicrosoftAzureActiveDirectoryApi.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/microsoftazureactivedirectory/BaseMicrosoftAzureActiveDirectoryApi.java
@@ -31,6 +31,11 @@ public abstract class BaseMicrosoftAzureActiveDirectoryApi extends DefaultApi20 
     }
 
     @Override
+    public String getDeviceAuthorizationUrl() {
+        return MSFT_LOGIN_URL + tenant + OAUTH_2 + getEndpointVersionPath() + "/devicecode";
+    }
+
+    @Override
     public ClientAuthentication getClientAuthentication() {
         return RequestBodyAuthenticationScheme.instance();
     }
--- a/scribejava-core/src/main/java/com/github/scribejava/core/oauth2/OAuth2Error.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/oauth2/OAuth2Error.java
@@ -1,5 +1,10 @@
 package com.github.scribejava.core.oauth2;
 
+import java.util.EnumSet;
+import java.util.Set;
+
+import static java.util.Collections.unmodifiableSet;
+
 public enum OAuth2Error {
     /**
      * @see <a href="https://tools.ietf.org/html/rfc6749#section-4.1.2.1">RFC 6749, 4.1.2.1 Error Response</a>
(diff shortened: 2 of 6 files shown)


scribejava-core/src/main/java/com/github/scribejava/core/oauth2/OAuth2Error.java
```diff
@@ -67,7 +72,29 @@
...
+     * @see <a href="https://tools.ietf.org/html/rfc8628#section-3.5">rfc8628#section-3.5</a>
+     */
+    EXPIRED_TOKEN("expired_token"),
+    ;
+
+    /**
+     * Unlike {@link #values()} which creates a new array every time, this always
+     * returns the same immutable object.
+     */
+    public static final Set<OAuth2Error> VALUES = unmodifiableSet(EnumSet.allOf(OAuth2Error.class));
```
kullfar, line 97: ...but a bit more memory consumption. CPU vs memory ;-)
ok, I will leave your variant.
except for unmodifiableSet. unneeded wrapper, it's better to make this member private (I have done this)

scribejava-apis/src/main/java/com/github/scribejava/apis/microsoftazureactivedirectory/BaseMicrosoftAzureActiveDirectoryApi.java
```diff
@@ -30,6 +30,11 @@ protected String getAuthorizationBaseUrl() {
         return MSFT_LOGIN_URL + tenant + OAUTH_2 + getEndpointVersionPath() + "/authorize";
     }
 
+    @Override
+    public String getDeviceAuthorizationUrl() {
+        return MSFT_LOGIN_URL + tenant + OAUTH_2 + getEndpointVersionPath() + "/devicecode";
+    }
```
kullfar, line 36: I have added Google implementation to the master, because I can test it. I have app and user.
Provide me with MS test app credentials, test user credentials, please, so I can test them before merging to the master.

Example as https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Google20DeviceAuthorizationGrantExample.java is needed as well, but it's easy to create (almost copy-paste).

kullfar:
Hi, I have merged this Pull Request with my additional changes to the master!
Thank a lot for this starting!

scribejava-core/src/main/java/com/github/scribejava/core/oauth2/OAuth2Error.java
rebarbora-mckvak: The old variant actually uses more memory and more CPU and copy and GC the array every time values() is called. The wrapper is needed when the field is public, which IMHO should be to let others use it. The wrapper has zero impact on performance.

kullfar: I mean, old variant uses memory only when it's invoked, but in cold state it doesn't need memory at all.

As for public members, I think it can break encapsulation. In case someone will need it outside, we will return to this discussion.

p.s. And for zero impact, I can't agree, at least at application startup, but this impact is soo small, so it's not the subject to discuss ;-)

rebarbora-mckvak: How an immutable field can break encapsulation?

kullfar: *public* member breaks encapsulation, not immutable field

rebarbora-mckvak: Could you please tell me what you mean by "it breaks encapsulation"? I would really like to learn about how to improve my code.

kullfar: you can use google ;-) for example first link
https://www.javatpoint.com/encapsulation

it's a good practice to not have a public member, but to give access to them via public methods. So you can hide real implementations inside the class and change it whenever you need.

rebarbora-mckvak: I understand encapsulation in the context of an object. I totally agree it is very good to have API that prevents programmers from changing internal state of an object freely, possibly making the state invalid. However this field is not part of any object state, it is a constant and I do not think it is a good or common practice to access constants using methods.

kullfar: Yes, you are right, there is no need for access methods for constants, but... to be honest, this member is not a constant by the sense. It's a constant because of modifiers "public static final". Actually this "constant" is a just internal optimisation.
Anyway, I think this discussion is more theoretical and about coding styles. Let's not spend our time on that.
In case of any practical problems with this code, we will return to this question.

PR #960, opened 2020-04-17
update for new path AuthbaseUrl

fquiroz:
GET https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id={client_id}&scope={scope}
    &response_type=token&redirect_uri={redirect_uri}


kullfar:
Hi, you should use https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/main/java/com/github/scribejava/apis/MicrosoftAzureActiveDirectory20Api.java#L46 instead of editing base parent class.

PR #962, opened 2020-04-22
Http headers QoL changes

mdindoffer:
This PR adds a quality of life change, that allows to add a map of HTTP headers at once.
It also properly encapsulates the Headers and Parameters maps of the OAuthRequest to not leak mutable collections.

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
@@ -101,6 +101,15 @@ public class OAuthRequest {
     }
 
     /**
+     * Add a collection of HTTP Headers to the Request.
+     *
+     * @param headers a key-value map of HTTP headers
+     */
+    public void addHeaders(Map<String, String> headers) {
+        this.headers.putAll(headers);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


scribejava-core/src/main/java/com/github/scribejava/core/model/OAuthRequest.java
```diff
@@ -100,6 +101,15 @@ public void addHeader(String key, String value) {
...
     }
 
+    /**
+     * Add a collection of HTTP Headers to the Request.
+     *
+     * @param headers a key-value map of HTTP headers
+     */
+    public void addHeaders(Map<String, String> headers) {
+        this.headers.putAll(headers);
+    }
```
kullfar, line 111: Hi, mdindoffer! Thanks for this Pull Request, I will add this method along with unit test for it.

As for Collections.unmodifiableMap - I don't think, we really need it, and, second, it can break current usages of this methods.

kullfar: Ah, no.
I remembered, why this method is still not here. Because, headers are not encapsulated, no need to wrap all methods of this Set. Let the user invoke getHeaders().putAll() or any other Set method.
And I don't see any reasons to make this encapsulation for the moment.

PR #978, opened 2020-08-13
Test improvement: removed exception handling (test smell)

gvma:
This is a test refactoring.

**Problem:**
The Exception Handling test smell occurs when the explicit passing or failing of a test method is dependent on the production method throwing an exception.

**Solution:**
The use of JUnit's exception handling to automatically pass/fail the test instead of writing custom exception handling code or throwing an exception is encouraged.

**Result:**
_Before:_
`try {
            future.get();
            fail();
} catch (ExecutionException expected) {
            // expected
}`

_After:_
`@Test(expected = ExecutionException.class) {
            ...
}`

### DIFF
--- a/scribejava-httpclient-okhttp/src/test/java/com/github/scribejava/httpclient/okhttp/OAuthAsyncCompletionHandlerTest.java
+++ b/scribejava-httpclient-okhttp/src/test/java/com/github/scribejava/httpclient/okhttp/OAuthAsyncCompletionHandlerTest.java
@@ -87,7 +87,7 @@ public class OAuthAsyncCompletionHandlerTest {
         assertEquals("All good", future.get());
     }
 
-    @Test
+    @Test(expected = ExecutionException.class)
     public void shouldReleaseLatchOnIOException() throws Exception {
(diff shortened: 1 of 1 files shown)


echorebel:
nice, i learned something

kullfar:
Hi, thanks for pointing this. But your way has one disadvantage I do not like -  the test will not fail if ExecutionException is thrown somewhere before future.get().

And if you want to analyze the Exception like here 

```
        try (Response response = error(responseBody)) {
            extractor.extract(response);
            fail();
        } catch (OAuth2AccessTokenErrorResponse oaer) {
            assertEquals(OAuth2Error.INVALID_GRANT, oaer.getError());
            assertEquals("unknown, invalid, or expired refresh token", oaer.getErrorDescription());
        }
```
https://github.com/scribejava/scribejava/blob/master/scribejava-core/src/test/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractorTest.java#L90

it's better to use ExpectedException approach.

But junit >=4.13 has nice assertThrows method. I used it. https://github.com/scribejava/scribejava/commit/cd20f57ff1554a6d23a85c3bd5abadb4032956a6


Anyway thx for the code improvement ;-)

kullfar:
Code is in the master, so I'm closing the Pull Request

PR #979, opened 2020-08-21
Fixes for use with Microsoft Graph

plutext:
With these changes, it is possible to use Microsoft Graph to upload/convert/delete a file, following https://medium.com/medialesson/convert-files-to-pdf-using-microsoft-graph-azure-functions-20bc84d2adc4

getAccessTokenEndpoint = MSFT_LOGIN_URL + tenant + OAUTH_2 + getEndpointVersionPath() + "/token";

For it to work for me, I have to have an empty EndpointVersionPath.

It is also necessary to pass a resource, to get the audience claim correct.   I have hard coded this, but if you process this PR, you'll want to handle this differently.

### DIFF
--- a/scribejava-core/src/main/java/com/github/scribejava/core/httpclient/jdk/JDKHttpClient.java
+++ b/scribejava-core/src/main/java/com/github/scribejava/core/httpclient/jdk/JDKHttpClient.java
@@ -189,6 +189,7 @@ public class JDKHttpClient implements HttpClient {
     }
 
     private static void addBody(HttpURLConnection connection, byte[] content, boolean requiresBody) throws IOException {
+    	if (!requiresBody && content==null) return;
         final int contentLength = content.length;
         if (requiresBody || contentLength > 0) {
             final OutputStream outputStream = prepareConnectionForBodyAndGetOutputStream(connection, contentLength);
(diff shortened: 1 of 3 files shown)


kullfar:
Hi, @plutext , Do you try to use https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/main/java/com/github/scribejava/apis/MicrosoftAzureActiveDirectoryApi.java ?
example is here https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/MicrosoftAzureActiveDirectoryExample.java

scribejava-core/src/main/java/com/github/scribejava/core/httpclient/jdk/JDKHttpClient.java
```diff
@@ -189,6 +189,7 @@ private static void addHeaders(HttpURLConnection connection, Map<String, String>
     }
 
     private static void addBody(HttpURLConnection connection, byte[] content, boolean requiresBody) throws IOException {
+    	if (!requiresBody && content==null) return;
```
kullfar, line 192: Could you show when can content be null?

PR #988, opened 2020-11-30
parameterize MS_LOGIN_URL

kimim:
Because Azure Active Directory supports four kind of LOGIN_URL (see: [Azure AD authentication endpoints](https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud#azure-ad-authentication-endpoints)):

- https://login.chinacloudapi.cn/
- https://login.microsoftonline.com/
- https://login.microsoftonline.de/
- https://login.microsoftonline.us/

So, we need to make MS_LOGIN_URL a constructor parameter.


srvrguy:
Is this a good candidate to get added to this code? I'd like to use this library, but basically have to make my own custom solution to supporting national cloud login. If there is something that needs to be changed, some feedback would be good so someone else can try submitting a PR or the original contributor could provide a fix to get this integrated.

gellweiler:
I also have a customer that wants to connect to a "national" cloud. It's off course relative easy to extend the class and override the url but I think it would be a good addition to the Scribejava core. So I second this.

PR #1001, opened 2021-02-26
Added OAuth2 support and examples for Twitter API

xiangyao1989:
1. Introduced OAuth2.0 config for Twitter API - TwitterApi20.java (Beta).
2. Created example of hitting Twitter APIs with OAuth2.0 PKCE - Twitter20WithPKCEExample.java.


scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Twitter20WithPKCEExample.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String NETWORK_NAME = "Twitter";
+    private static final String PROTECTED_RESOURCE_URL = "https://api.twitter.com/2/tweets?ids=1261326399320715264,1278347468690915330";
+
+    private Twitter20WithPKCEExample() {
+    }
+
+    @SuppressWarnings("PMD.SystemPrintln")
+    public static void main(String... args) throws IOException, InterruptedException, ExecutionException {
+        final String clientId = "CLIENT_ID"; // replace these with your client id
+        final String state = "secret" + new Random().nextInt(999_999);
```
lillieMaiBauer, line 33: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once).
This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.
If it is important that the generated Random numbers not be guessable, you *must* not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).

horyu1234:
@xiangyao1989 Any further news on this work?

peternees:
I had a look to the endpoints in TwitterApi20.java, and I don't think they are correct, if I compare with [https://developer.twitter.com/en/docs/authentication/oauth-1-0a/obtaining-user-access-tokens](url)

PR #1008, opened 2021-05-14
Added Etrade api.

matias-favale:
Tested this locally successfully.


BR-Nicholas-Prado:
Perhaps this worked at the time, but Etrade currently rejects this oauth attempt.

### steps
1. used EtradeApiExample as the main class of a new project
1. changed the key and secret to match my sandbox key and secret
1. ran with gradle
1. pasted the authorize application url into a browser
1. provided my etrade credentials to that web page within 60 seconds

### expected behavior
etrade page asks for me to consent to the application

### observed behavior
etrade page complains that a login delay or other problem prevented authentication

I know that the test example class uses the sandbox api rather than the production api because the maven checkstyle complained when I tried to build a jar from the fork. I know that my credentials are correct because I use a password manager to log in to etrade and used the same. I know that the api key and secret are correct because I used etrade's example code to perform oauth with it and that showed their terms page.

For whatever reason, this pr does not correctly authenticate with etrade, at this time.

PR #1012, opened 2021-06-20
new Api for EveOnline

Halderian:


M-F-K:
I know this is an old PR ( and I'm not the owner of this project) But anyway, this PR seems to at least be missing two things:

1) The example file ( se the other integration for naming and so on)
2) A line in the Read.me file mentioning this api.

also filling out the description would be nice =)

Issue #1016, opened 2021-07-20
Support Apache HttpClient 5


davidkron:
It would be great if the current major version of Apache HttpClient would be supported, which is 5:
https://hc.apache.org/httpcomponents-client-5.1.x/index.html

kullfar:
Hi. It's not an issue.
It's better to submit Pull Requests with such enhances ;-)

davidkron:
You are right, sorry for misusing the issue tracker for this request.

Since we are currently using HttpClient 5 in our project and want to use Scribe as well, I think I will have to write an implementation anyway. So I will create a pull request if I have something working.

PR #1017, opened 2021-07-20
Apache HttpClient 5 integration

davidkron:
Based on the ticket I created:
https://github.com/scribejava/scribejava/issues/1016

I implemented this integration of Apache HttpClient 5.1 for Scribe. My code is more or less a copy of the existing HttpClient 4.x implementation (https://github.com/scribejava/scribejava/tree/master/scribejava-httpclient-apache) with the necessary modifications to dependencies and the package structure that changed with httpclient5.

What I am not 100% sure about, is the whole asynchronous stuff from HttpClient as I primarily have experience with the synchronous API. The whole asynchronous API in HttpClient 5 doesn't seem to be using the same approach anymore as before with the HttpResponse containing also the HttpEntity/InputStream. But there are some compatibilty classes available, so that's why I choose to stay as close to the previous implementation as possible, especially since the Scribe API primarily is using InputStream.

I used and tested the same implementation also in our project, but as I just implemented it, I don't yet have any experience over a longer period of usage.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
         <module>scribejava-httpclient-ning</module>
         <module>scribejava-httpclient-okhttp</module>
         <module>scribejava-httpclient-apache</module>
+        <module>scribejava-httpclient-apache5</module>
         <module>scribejava-httpclient-armeria</module>
     </modules>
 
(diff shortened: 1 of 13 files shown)


dizzzz:
Looks like a valuable addition to me....

dizzzz:
Can I offer some help to have this pulled in? Is there a procedure for this?

darktyco:
Definitely a valuable feature to have in scribe since Spring 6 / Spring Boot 3 have moved on to HttpClient 5

Issue #1023, opened 2021-10-08
SlackJsonTokenExtractor bug when access_token is not found in response body 


jaxony:
Hi there! I'm in the process of migrating Slack V1 App to V2 but ran into an error.

The code here expects that `access_token` is required, but in fact with Slack V2, `access_token` (meaning the "bot token") can be missing from the response object if the client only requested for "user-scoped permissions". 

https://github.com/scribejava/scribejava/blob/124745961e999ccede90e2348c6012f8d335eb8a/scribejava-core/src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java#L90

A proposed solution is to override the `OAuth2AccessTokenJsonExtractor#createToken(String rawResponse)` method in `SlackJsonTokenExtractor`, making the access token an optional parameter instead of a required one.

Happy to do the PR for this if the owner is happy with the proposed solution 😄

PR #1037, opened 2022-04-28
SlackJsonTokenExtractor handle null access_token

peterp-canva:
Issue: https://github.com/scribejava/scribejava/issues/1023
Slack does not always provide the bot access token, if you don't ask for those scopes.
Example:
- Requesting: slack oauth `&scope=&user_scope=users.profile%3Aread`
- Response
```json
{"ok":true,"app_id":"AA","authed_user":{"id":"UID","scope":"channels:read,groups:read,users.profile:read,files:write:user","access_token":"xoxp-1234-etc","token_type":"user"},"team":{"id":"TId","name":"TeamName"},"enterprise":{"id":"EId","name":"EName"},"is_enterprise_install":false}
```

Error - This leads to:
```
com.github.scribejava.core.exceptions.OAuthException: Response body is incorrect. Can't extract a 'access_token' from this: '{"ok":true,"app_id":"AA",... <as above>'
```

### Solution:
Downgrade "access_token" to not be extractRequiredParameter, by overriding the implementation.
https://github.com/scribejava/scribejava/blob/9c8e4c12df64fc15a9d7fb0335156e918977a39a/scribejava-core/src/main/java/com/github/scribejava/core/extractors/OAuth2AccessTokenJsonExtractor.java#L90

### DIFF
--- a/scribejava-apis/src/main/java/com/github/scribejava/apis/slack/SlackJsonTokenExtractor.java
+++ b/scribejava-apis/src/main/java/com/github/scribejava/apis/slack/SlackJsonTokenExtractor.java
@@ -2,6 +2,12 @@ package com.github.scribejava.apis.slack;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.github.scribejava.core.extractors.OAuth2AccessTokenJsonExtractor;
+import com.github.scribejava.core.model.OAuth2AccessToken;
+import com.github.scribejava.core.model.OAuthConstants;
+import com.github.scribejava.core.model.Response;
+import com.github.scribejava.core.utils.Preconditions;
+
+import java.io.IOException;
 
 public class SlackJsonTokenExtractor extends OAuth2AccessTokenJsonExtractor {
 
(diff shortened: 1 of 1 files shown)


scribejava-apis/src/main/java/com/github/scribejava/apis/slack/SlackJsonTokenExtractor.java
```diff
@@ -17,6 +23,37 @@ public static SlackJsonTokenExtractor instance() {
...
+        final JsonNode response = OBJECT_MAPPER.readTree(rawResponse);
+
+        final JsonNode expiresInNode = response.get("expires_in");
+        final JsonNode refreshToken = response.get(OAuthConstants.REFRESH_TOKEN);
+        final JsonNode scope = response.get(OAuthConstants.SCOPE);
+        final JsonNode tokenType = response.get("token_type");
+        final JsonNode globalAccessToken = response.get(OAuthConstants.ACCESS_TOKEN);
+
+        return createToken( //
+                globalAccessToken == null ? "" : globalAccessToken.asText(), // Avoid "access_token can't be null"
```
peterp-canva, line 50: Note - the underlying OAuth2AccessToken (for SlackOAuth2AccessToken->super()) has a Precondition for "access_token can't be null", so here it is provided with an empty string.
The tokenType will be null, so it is easy to deduce whether the string is meaningful.

peterp-canva:
Closing as this has gone stale / to clear open PRs.

PR #1040, opened 2022-06-04
[readme] Change variable naming in example

tipsy:
This is a more common convention, which is also used in your own examples: https://github.com/scribejava/scribejava/blob/master/scribejava-apis/src/test/java/com/github/scribejava/apis/examples/Google20Example.java#L27-L35

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,8 @@
 Who said OAuth/OAuth2 was difficult? Configuring ScribeJava is __so easy your grandma can do it__! check it out:
 
 ```java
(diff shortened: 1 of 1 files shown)


tipsy:
Well, I guess this is not going to happen.

kullfar:
Yeah.. Why not? You're absolutely right. client_id and client_secret naming are correct namings according to RFC https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.1

So, let it be so in readme.md as well

ps. Sorry for a long delay ;-)

Issue #995, opened 2021-01-19
Keycloak: Different context root


virtualmarc:
By default Keycloak is deployed at a context root /auth which can be changed in the configuration.
Out Keycloak Server is deployed in the context root / which makes scribejava generate wrong URL's.

The problem is in
https://github.com/scribejava/scribejava/blob/cd20f57ff1554a6d23a85c3bd5abadb4032956a6/scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java#L41

where `auth/` is hardcoded.

In other Clients/Keycloak Lib's the context root has to be set in the url.

danielFesenmeyer:
Stumpled across the same issue. The "auth/" should just be removed - it still can be passed as part of the baseUrl, when it is needed.

scrhall:
I too have encountered this problem. For now the only way I have found is to copy the API to a local class and remove that hardcoded

jocko-angi:
Hopefully I'm bumping this.  The PR is ready, our org could use this fix.

dawnstrider:
Was experimenting with this lib today and had the exact same issue. Please remove the hardcoded "auth/" prefix from the baseurl-composition.

PR #1044, opened 2022-07-20
Configurable context root for keycloak API

virtualmarc:
Fixes issue #995


jocko-angi:
@virtualmarc -- How can we get movement on merging this PR.   My org has a need for this fix, and it would definitely keep the configuration in parity with the current default configs for Keycloak.

jocko-angi:
@kullfar -- Do you know if there is a timeline for this fix?  Or how we can inspire movement on it?  Thanks!

Issue #1060, opened 2023-07-19
Add Zendesk Oauth 2.0 Support


nis130:
Add support for Zendesk Oauth 2.0 out of the box in the library with the example.

Ref: https://support.zendesk.com/hc/en-us/articles/4408845965210-Using-OAuth-authentication-with-your-application

PR #1061, opened 2023-07-19
add support for zendesk oauth 2.0

nis130:
Zendesk Oauth Documentation: https://support.zendesk.com/hc/en-us/articles/4408845965210-Using-OAuth-authentication-with-your-application

fixes: #1060


nis130:
@kullfar

PR #1069, opened 2024-03-21
[Issue#1068] KeycloakApi doesn't work with Keycloak version >= 17

Julesfpollender:
Fixes for issue#1068


scribejava-apis/src/main/java/com/github/scribejava/apis/KeycloakApi.java
```diff
@@ -51,6 +53,11 @@ protected String getAuthorizationBaseUrl() {
         return baseUrlWithRealm + "/protocol/openid-connect/auth";
     }
 
+    @Override
+    public ClientAuthentication getClientAuthentication() {
+        return RequestBodyAuthenticationScheme.instance();
+    }
```
rvillablanca: Why was this change required?

Julesfpollender: I don't think it is required. I will remove it

